PR #2, opened 2022-06-02
Collating ideas from the meeting 5/31/2022 Meeting

skilesare:
I threw together this from the comments on the last meeting.  Everything is open for comment or criticism.  A couple of notes:

- I think we are going to have to support #pending return types especially when we get eth and BTC integration, so I updated the variant there.
- #noop is there in case a ledger wants to not charge a fee for an update call of some kind.  All should probably have a fee of some kind to prevent spamming.
- I've obviously updated this with namespacing because I think it is a good idea. All the functions end in _ic1. What if you are implementing two standards on one canister that use decimals?  For example, a ledger that supports trading some kind of trade interface that has a decimals() call that returns the number of supported decimals in the price field.  Won't someone think of the decimals?
- I've put a single principal to get the notification in there on transactions. If you leave it blank then no transaction. To do this we all need to agree on a standard function and payload. I haven't really looked at the block payload yet, but that probably deserves some debate as well.
- Approving has a fee and suggests moving funds to a separate sub-account, although I think that could technically be an implementation detail, but I think it should be a best practice.
- I considered changing the TOs and FROMS to a {#subaccount: {principal: Principal; subaccount: Blob;}; #account: Text} format, but held off for now as I'm not sure it is necessary.  Once we can dump subnet blocks people will be sifting through this info and I guess we should consider if we want that to be easy or hard.
- Before I forget, we need a batch send/approve/transferFrom, but I haven't added it yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,21 @@
 
 The ICRC-1 is a standard for Fungible Tokens on the [Internet Computer](https://internetcomputer.org).
 
+## Data structures
+
+```
+
+type TransactionResponse = {
+  #ok : { //index confirmed
+   #index: nat64;
+   #hash : Blob; //to allow for hash blobs
+  }; 
+  #err: TransactionError;
+  #pending: Blob; //hash if pending
+  #noop; //to support "free" operations
+
+```
+
 ## Methods
 
 ### name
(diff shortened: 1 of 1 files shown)


MarioDfinity:
@skilesare I'm closing this as we moved on from the proposals

PR #3, opened 2022-06-02
add notify

MarioDfinity:


MarioDfinity:
I'm closing this because notify is not going to be in the standard.

PR #6, opened 2022-06-05
Notify methods descriptions

blutooth:


MarioDfinity:
I'm closing this as notify is not going to be in the standard.

PR #7, opened 2022-06-08
[improve] did and readme

witter-deland:
1. **Why not use (to_principal, to_subaccount) as the receiver parameter?**

When transferring token, the only thing the user can provide is the principal or accountid, subaccount is too difficult for ordinary users, so the transfer parameters should not use (to_principal, to_subaccount)

This is why the transfer parameter and the balanceOf parameter are modified

2. **Why to(the receiver) use text？** 
It can support the user to use either the principal or accountid string format

3. **Does the support principal and account id conflict with the definition of account?**
It won't. When the user provides the principal, the token standard implementation can internally convert to the default accountid for the user (subaccount = 0)

4. **Using to:text, if the wrong format of principal/account id is provided, what should token standard do?**
Check it internally first.
 
5. **Why do decimals use nat8?** 
Do your decimal places go beyond nat8? I don't think so.

6. **Why change totalSupply return to nat?**
nat32 is too small for many scenarios

### DIFF
--- a/ICRC-1.did
+++ b/ICRC-1.did
@@ -2,9 +2,8 @@ type SubAccount = blob;
 
 type TransferArgs = record {
     from_subaccount: opt SubAccount;
-    to_principal: principal;
-    to_subaccount: opt SubAccount;
-    amount: nat64;
+    to: text;
+    amount: nat;
 };
 
 type TransferError = variant {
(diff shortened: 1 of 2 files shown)


MarioDfinity:
Thanks for the PR. My thoughts:

> Why not use (to_principal, to_subaccount) as the receiver parameter?
> When transferring token, the only thing the user can provide is the principal or accountid, subaccount is too difficult for > ordinary users, so the transfer parameters should not use (to_principal, to_subaccount)

The working group that works for to make the standard decided to go with plain principal and subaccount. The standard should not target users as in real scenarios the arguments are compiled by either a canister, e.g. a dex, or a wallet app.

Note that there is no ICP-like accountid anymore. It's plain principal and subaccount. Account ids could end up in an alternative standard for a secure ledger but the working group leaned toward having transparent principal and subaccount because the security provided by account ids is limited.

> Why do decimals use nat8?
> Do your decimal places go beyond nat8? I don't think so.
>
> Why change totalSupply return to nat?
> nat32 is too small for many scenarios

I hope we will be able to discuss this details too in a working group session. The current API for decimals and totalSupply is still to be defined.

PR #13, opened 2022-07-04
Proposal: supportedStandards endpoint

roman-kashitsyn:
This change proposes to include an endpoint that will allow an
implementation to advertise that it supports other standards that
we might introduce later.


MarioDfinity:
Let's discuss about this during the 12 Jul 2022 working group session.

MarioDfinity (approved):

PR #19, opened 2022-07-14
memo blob -> nat

MarioDfinity:


ggreif (commented):
I'd suggest `nat64`. `nat` is defined as infinite precision. Do you have any requirement for more than 64 bits?

MarioDfinity:
No we don't. We use nat for amount but memo is something different and having infinite precision for memo may be problematic. I switched to nat64. Thanks @ggreif .

PR #26, opened 2022-07-20
Account and SubAccount types

lastmjs:
I think `SubAccount` should be specified, and it would be nice to have an `Account` type instead of a separate principal and optional SubAccount everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ A `principal` can have multiple accounts. Each account of a `principal` is ident
 
 The account identified by the subaccount with all bytes set to 0 is the _default account_ of the `principal`.
 
+```
+type SubAccount = nat;
+type Account = record { principal; opt SubAccount; };
+```
+
 ## Methods
 
 ### icrc1_name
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,11 @@ A `principal` can have multiple accounts. Each account of a `principal` is ident
 
 The account identified by the subaccount with all bytes set to 0 is the _default account_ of the `principal`.
 
+```
+type SubAccount = nat;
+type Account = record { principal; opt SubAccount; };
```
roman-kashitsyn, line 15: Shouldn't it be
```suggestion
type SubAccount = blob;
type Account = record { of: principal; subaccount : opt SubAccount; };
```
?

MarioDfinity: I like the usage of `of` as field name because it explains that the principal is the owner.

lastmjs:
I'm not sure I can sign the CLA, as part of signing it it asks me to state that I'm not a contractor of DFINITY, but Demergent Labs receives funding from DFINITY.

lastmjs:
I've updated the pull request with your suggestions

MarioDfinity:
Thank you @lastmjs !

README.md
blutooth: type SubAccount = vec nat8;

rather than blob?

PR #27, opened 2022-07-20
feat: draft of a reference implementation

roman-kashitsyn:
This change introduces a draft of the reference implementation for the standard.
The implementation is not well tested yet but appears to work for basic flows.


jzxchiang1:
Drive-by comment, but the `totalSupply` method implementation seems rather inefficient, especially when compared with the Rust ICRC-1 implementation of that same method: https://github.com/dfinity/ic/blob/2cac973e810645f915e8fd0019ad2586258d52d9/rs/rosetta-api/ledger_core/src/balances.rs#L174

PR #29, opened 2022-07-25
Proposal: icrc1_minting_account endpoint

roman-kashitsyn:
This change introduces the notion of a minting account and adds
an endpoint that returns the current minting account of the ledger.


ICRC-1.did
```diff
@@ -39,7 +43,8 @@ service : {
     icrc1_symbol : () -> (text) query;
     icrc1_decimals : () -> (nat8) query;
     icrc1_total_supply : () -> (nat) query;
-    icrc1_balance_of : (record { of: principal; subaccount: opt SubAccount; }) -> (nat) query;
-    icrc1_transfer : (TransferArgs) -> (variant { Ok: nat; Err: TransferError; });
+    ircr1_minting_acount : () -> (Account) query;
```
lastmjs, line 46: Typo: `ircr1_minting_acount` should be `icrc1_minting_acount

README.md
```diff
@@ -54,12 +54,20 @@ icrc1_metadata : () -> (vec { record { text; Value } }) query;
...
 ```
 icrc1_total_supply : () -> (nat) query;
 ```
 
+### icrc1_minting_account
+
+Returns the [minting account](#minting_account) for this ledger.
+
+```
+icrc1_minting_account : () -> (Account) query;
```
lastmjs, line 68: How is the minting account set? Should the service have a parameter for the minting account?

lastmjs (changes requested):

README.md
lastmjs: We discussed this during the WG call today, I understand why it's left up to the implementation how the minting account is set.

lastmjs:
I'm starting to think that the minting account should be an extension to ICRC-1. Not all tokens currently have a minting account, and I foresee not all tokens needing/wanting a minting account in the future.

dansteren:
> I'm starting to think that the minting account should be an extension to ICRC-1. Not all tokens currently have a minting account, and I foresee not all tokens needing/wanting a minting account in the future.

I agree with Jordan and would prefer we leave this to an extension. If however we are in the minority, I suggest we at least make the return type an `opt Account` to accommodate those scenarios.

ICRC-1.did
roman-kashitsyn, line 46:
```suggestion
    icrc1_minting_account : () -> (Account) query;
```

roman-kashitsyn: Thanks for catching!

roman-kashitsyn:
> I'm starting to think that the minting account should be an extension to ICRC-1. Not all tokens currently have a minting account, and I foresee not all tokens needing/wanting a minting account in the future.
> 
> It's a valid use-case to want a fixed supply of tokens to exist at launch.

That's a good point.

Maybe we could move this information into metadata?

We could add a new Value variant:
```
type Value = variant { ... Principal : principal };
```
and return `record { "icrc1:minting_account"; variant { Principal = ... } }` metadata entry for ledgers that have a minting account.

roman-kashitsyn:
The working group discussed this feature today.
The agreement is to add the `icrc1_minting_account` with optional result and extend it later with the corresponding metadata entry once we define textual representation for accounts.

PR #32, opened 2022-07-26
Proposal: Rename principal in `Account` type

dansteren:
The `Account` type is a combination of a principal and a subaccount. The subaccount was correctly called "subaccount" but the principal was called "of" for some reason. In the context of the `icrc1_balance_of` method, this makes sense. However, in the context of `icrc2_transfer` it does not.

In the future we may implement even more methods where "of" doesn't make sense as the name of the principal. By separating the usage from the definition, `Account` will remain flexible for all these use cases.

Additionally, when #26 was merged in we didn't update the did file. This takes care of that.

### DIFF
--- a/ICRC-1.did
+++ b/ICRC-1.did
@@ -4,12 +4,16 @@ type Timestamp = nat64;
 // Number of nanoseconds between two [Timestamp]service
 type Duration = nat64;
 
-type SubAccount = blob;
+type Subaccount = blob;
+
+type Account = record {
+    of: principal;
+    subaccount: opt Subaccount;
+};
 
 type TransferArgs = record {
-    from_subaccount: opt SubAccount;
-    to_principal: principal;
-    to_subaccount: opt SubAccount;
+    from_subaccount: opt Subaccount;
+    to: Account;
     amount: nat;
     fee: opt nat;
     memo: opt nat64;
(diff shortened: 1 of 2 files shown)


roman-kashitsyn:
Thanks, @dansteren!

>  The subaccount was correctly called "subaccount" but the principal was called "of" for some reason.

I guess I'm responsible for calling the field `of`.
The only reason for that spelling is that `Account { of : principal }` looks irresistibly neat to me (reads very naturally, isn't it?). The `of` part indicates the owner of the account. That spelling reminds me a bit of Objective-C naming conventions that I find very readable. I now realize that it doesn't look good in Motoko, where we don't specify the type name when constructing a record.

Since you are not the first one who is confused with this naming, we should consider changing it.
`record { principal : principal }` also reads well, both in `transfer` and in `balance_of` context:

```
transfer({ to = { principal = myPrincipal } ... })
balance_of({ principal = myPrincipal })
```

roman-kashitsyn (approved):

README.md
```diff
@@ -12,7 +12,7 @@ The account identified by the subaccount with all bytes set to 0 is the _default
 
 ```
 type Subaccount = blob;
-type Account = record { of: principal; subaccount: opt Subaccount; };
+type Account = record { principal: principal; subaccount: opt Subaccount; };
```
roman-kashitsyn, line 15:
```suggestion
type Account = record { "principal": principal; subaccount: opt Subaccount; };
```

ICRC-1.did
```diff
@@ -4,12 +4,16 @@ type Timestamp = nat64;
 // Number of nanoseconds between two [Timestamp]service
 type Duration = nat64;
 
-type SubAccount = blob;
+type Subaccount = blob;
+
+type Account = record {
+    principal: principal;
```
roman-kashitsyn, line 10: The candid file doesn't pass validation. I think we need to put the field name into quotes because it's a reserved symbol.
```suggestion
    "principal": principal;
```

README.md
dansteren: :+1: Nice catch!

ICRC-1.did
dansteren: Ah, I didn't even think of running it through didc. I've updated the type to have quotes and ran didc locally and it all passed. This should be good now!

Issue #30, opened 2022-07-26
About data structures


iclighthouse:
1. Account
```
type Account = record { of: principal; subaccount: opt Subaccount; };
icrc1_balance_of : (record { of: principal; subaccount: opt SubAccount; }) -> (nat) query;
```
`type Account = blob` might be a better choice.
It is only necessary to specify the SubAccount when calculating Caller's Account.

2. Timestamp
```
type TransferArgs = record {
    from_subaccount: opt Subaccount;
    to: Account;
    amount: nat;
    fee: opt nat;
    memo: opt nat64;
    created_at_time: opt Timestamp;
};
```
If `Timestamp` is of type Nat64, which means nanoseconds, then after 500 years the data will overflow.

3. TransferError
```
type TransferError = variant {
    BadFee : record { expected_fee : nat };
    BadBurn : record { min_burn_amount : nat };
    InsufficientFunds : record { balance : nat };
    TooOld : record { allowed_window_nanos : Duration };
    CreatedInFuture;
    Duplicate : record { duplicate_of : nat };
    GenericError: text;
};
```
Token and ICP Ledger have different usage scenarios and require careful definition of the Error data structure.

roman-kashitsyn:
Hi @iclighthouse!

Thanks for the constructive feedback!

> type Account = blob might be a better choice.

What will the blob contain? Why do you think packing the principal and the subaccount into a blob might be a better choice?

> Token and ICP Ledger have different usage scenarios and require careful definition of the Error data structure.

I full-heartedly agree that we need to think hard about the error type definition.
I have a preliminary implementation of the standard in both Motoko and Rust, and the proposed error type seems to work well (except that `GenericError` should be `GenericError : record { error_code : nat; error_message : text }`; otherwise, it's no better than a reject).
Which error variants do you think we should remove or add?

iclighthouse:
> What will the blob contain? Why do you think packing the principal and the subaccount into a blob might be a better choice?
"principal+sub" => account-id
Account-id can be calculated outside the canister, unless the caller role. So, except for the caller, other roles enter the account-id directly. ICP Ledger is implemented in the same way.

`GenericError : record { error_code : nat; error_message : text }` I agree that this is a better solution.
The difference between Tokens and ICP is that Tokens has multiple implementations of different entities and ICP has only one on IC.

In addition, I make one more suggestion.
`icrc1_metadata : () -> (vec { record { text; Value } }) query;`
[...]

PR #33, opened 2022-07-26
Proposal: fee query and metadata

roman-kashitsyn:
This change introduces the icrc1_fee query call for obtaining the
transfer fee and the corresponding metadata entry as suggested in #30.


roman-kashitsyn:
The working group had a meeting today and unanimously accepted this feature.

PR #38, opened 2022-08-02
Proposal: larger memos

roman-kashitsyn:
This change proposes to extend memos from 8 bytes to a blob of at least
32 bytes in length.

This change has been discussed in the Working Group and suggested again in #30.


roman-kashitsyn:
The working group unanimously accepted this feature during the meeting today.

PR #39, opened 2022-08-02
TransferError - TemporarilyUnavailable

levifeldman:
For when a ledger wants to slow down the rate of the transfer calls. 

sample use case: to maintain a max-number of transactions in a single-canister's memory to check for a duplicate-transaction within a time-window.

### DIFF
--- a/ICRC-1.did
+++ b/ICRC-1.did
@@ -27,7 +27,7 @@ type TransferError = variant {
     TooOld : record { allowed_window_nanos : Duration };
     CreatedInFuture;
     Duplicate : record { duplicate_of : nat };
-    MaximumLoad;
+    TemporarilyUnavailable;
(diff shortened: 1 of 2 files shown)


ICRC-1.did
```diff
@@ -27,6 +27,7 @@ type TransferError = variant {
     TooOld : record { allowed_window_nanos : Duration };
     CreatedInFuture;
     Duplicate : record { duplicate_of : nat };
+    MaximumLoad;
```
roman-kashitsyn, line 30: Maybe `TemporarilyUnavailable` or `ServiceUnavailable` to draw on existing HTTP expertise?

```suggestion
    TemporarilyUnavailable;
```

levifeldman: sure

PR #43, opened 2022-08-03
align TransferError to 2022-08-02 wg decisions

MarioDfinity:


roman-kashitsyn:
Related change: https://github.com/dfinity/ICRC-1/pull/39
We should probably remove `Throttled`, `TemporarilyUnavailable` sounds better to me.

roman-kashitsyn (approved):

PR #55, opened 2022-08-29
feat: specify textual encoding

### LINKED ISSUE #47: Textual encoding for accounts
https://forum.dfinity.org/t/icrc-1-account-human-readable-format/14682

roman-kashitsyn:
This change proposes to use the Principal encoding for ICRC-1 accounts.

This design has a few attractive properties:

  1. All existing principals are valid ICRC-1 accounts out of the box.
  2. Most clients can decode principals already; adding support for ICRC-1 account encoding will be straightforward. I implemented a prototype in Motoko in a few lines of code.
  3. The checksum covers the principal and subaccount combination, providing good protection against copy-paste errors.
  4. For accounts with a non-empty subaccount, the encoding retains most of the digits of the original account. E.g., "4kydj-r**yaaa-aaaag-qaf7a-cai**" becomes "y7l7y-d**yaaa-aaaag-qaf7a-cai**aa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaa" when we add the default subaccount.
  

Closes #47.


README.md
```diff
@@ -217,6 +217,41 @@ variant { Err = variant { BadBurn = record { min_burn_amount = ... } } }
...
+Applications SHOULD encode accounts as follows:
+
+  1. The encoding of accounts that do not have a subaccount, such as `{ owner = principal "aaaaa-aa"; subaccount = null }`, is the encoding of the owner principal.
+  2. The encoding of accounts with a non-empty subaccount is the encoding of the concatenation of the owner principal bytes and the subaccount bytes.
+
+In pseudocode:
+
+```sml
+encodeAccount({ owner; subaccount = null      }) = Principal.toText(owner)
+encodeAccount({ owner; subaccount = opt bytes }) = Principal.toText(owner · bytes)
```
Dfinity-Bjoern, line 236: I don't think this is a good idea. The trailing byte of a principal determines its type, and appending an arbitrary byte string to it will create confusion.

roman-kashitsyn: Do you think we should add a new tag for such "extended" principals?
One property of that approach that we find very attractive is that each principal is a valid ICRC-1 account; I'd like to keep this property.

Dfinity-Bjoern: @bogwar convinced me that there is no ambiguity here, since the length of the encoded string is not a valid principal – principals have a bounded length of 29 bytes and sub-accounts are 32 bytes.

I'm still not a big fan of this since I still think it may create confusion on the user side (e.g. a subaccount of the anonymous principal `0x04` is visually almost the same as a self-authenticating principal), but I also don't see a better option for now.

timohanke: First of all I find the sheer length of it very unattractive from a perspective of user experience. The use case where all 32 bytes of the subaccount id space are actually used is niche. With this solution we are forcing everyone to deal with encodings of this length. There has to be a better way. Users who only use subaccounts 1,2,3 don't have to be bothered with encodings of this length (and with, to them, inexplicably many a's).

I still vote for `4kydj-ryaaa-aaaag-qaf7a-cai:1`, etc., i.e. having the two components exposed. If principals really have to be valid account ids (I am not totally convinced why) then `4kydj-ryaaa-aaaag-qaf7a-cai` can be allowed as a shortcut for `4kydj-ryaaa-aaaag-qaf7a-cai:0`.

bogwar: @timohanke your proposals are attractive since they look much nicer. But:

* I would imagine that a common use is for users to make payments to services -- in this case they most likely would have to deal with long subaccounts (if services embrace the payment flow where each payment goes into a specific subaccount). 

* they lack a checksum over the entire string. Wouldn't that be an issue with copy pasting? (Granted, the issue is somewhat alleviated since the transfer would still go to the intended principal, even if to the wrong subaccout)

timohanke: So let's say there are two different use cases, one where the subaccount is small (including the case when it is 0 and you when would like to not have to mention it at all) and the other one where it is large (20 bytes+). Isn't it then also ok to have two different encoding schemes? The encoding doesn't have to be unique. The first scheme can use the explicit version where the two components are visible. And that one is probably fine without a checksum. The second scheme can use a long homogeneous looking string. If the use cases are sufficiently different then it would be ok to have different encodings. Then you also have more freedom in the second scheme. E.g. property 1 doesn't have to embed as nicely anymore into the second scheme because it already does into the first scheme.

timohanke: As for the original proposal, I noticed that in order to achieve property 1) (that existing principals are valid account ids) there is a switch between two cases happening: if the length is <= 29 bytes then it is interpreted as a principal and the subaccount 0 is assumed, if the length is >= 33 bytes then the last 32 bytes are taken as the subaccount and all bytes before that are taken as the principal. If such a switch is ok then we can achieve variable length as follows:
- define a tag byte which differs from all existing principal types, e.g. 255 as existing types are 1-4,
- take the longest bytes prefix of the subaccount bytes that does not end in a zero byte,
- take the byte-length of that prefix and encode it into a byte (the "length byte")
- encode the account id as:
<principal bytes> || <prefix bytes> || <length byte> || <tag>
Then it is uniquely parsable, even if the overall length is <29 bytes, and distinguishable from a pure principal.

timohanke: Regarding the encoding, bech32 is a good alternative. It looks more homogenous (no dashes), less chance to confuse with a principal. The dashes are there to aid visual comparison of two principals which probably isn't relevant for account ids.

skilesare:
If we are able to find a text encoding that is principal like we have a much better chance of unifying the ecosystem. I’d lean that way, but understand the challenges.

README.md
timohanke: > We can specify in the IC interface spec that principals ending with bytes >= 128 aren't valid and will never be

That way you are occupying half of the entire space available in the type byte and taking it away from future extensions. I wouldn't do that only to save one byte in the overall length.

> For now, I still fix the length of the subaccount to be 32

I you want to do that then I would rather have you occupy two values in the type byte instead of 127. Define one for the fixed length and another one for the variable length. Especially if you want to leave the variable length for later anyway.

roman-kashitsyn: > But to be honest I don't understand why we need the fixed length (32 bytes) variant. If someone uses a collison resistant hash as subaccount id then sha-224 is popular. If someone uses a pre-image resistant hash as subaccount id then 20 bytes is popular. If someone uses another principal as subaccount id then it is also shorter than 32 bytes.

Enforcing a specific length allows for more efficient storage. 32 bytes is an upper limit because some want to use SHA-256 checksum as a subaccount. Everything else you can always pad with zeros.

roman-kashitsyn: > That way you are occupying half of the entire space available in the type byte and taking it away from future extensions. I wouldn't do that only to save one byte in the overall length.

Not sure we care much about the space: so far we have used only 4 tags, and one of them has never been used in practice. Even if we reserve 128 variants for variable-length, we still have 122 more tags to use. I think that's more than enough given the rate of tag usage...
But I don't have a strong opinion, <leb128 length><magic_byte> will work for me as well, it's just more work to implement.

timohanke: What I meant was that if someone has a shorter content in mind for a subaccount id, whether that's decimal 1,2,3,..., a 20 byte hash, another principal's bytes, a 28 byte hash or a 32 byte hash, you embed that into 32 bytes, i.e. pad with zeros if necessary and that is what you store. You store the full 32 bytes and that is also what the subaccount id is. The subaccount id _is_ the full 32 bytes. But now we are talking about the textual encoding of the 32 bytes long subaccount id to get a text string that you can copy-paste and send around. For that part I propose to define the encoding such that trailing zero bytes do not have to appear in the encoding. It does not affect storage but it does effect user experience. 

I think if all our account ids look like this `y7l7y-dyaaa-aaaag-qaf7a-caiaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaaa-aaaa`, if everyone is forced to have this length, then we are dead as a project.

README.md
```diff
@@ -217,6 +217,43 @@ variant { Err = variant { BadBurn = record { min_burn_amount = ... } } }
...
+
+Applications SHOULD encode accounts as follows:
+
+  1. The encoding of accounts that do not have a subaccount, such as `{ owner = principal "aaaaa-aa"; subaccount = null }`, is the encoding of the owner principal.
+  2. The encoding of accounts with a non-empty subaccount is the encoding of the concatenation of the owner principal bytes, the subaccount bytes, and an extra byte `A0`<sub>`16`</sub> (160 = 128 + 32).
+
+In pseudocode:
+
+```sml
+encodeAccount({ owner; subaccount = null      }) = Principal.toText(owner)
```
qti3e, line 235: what if a subaccount is filled with zeros (`[0;32]`) but not null? I believe that should also be considered null?

roman-kashitsyn: Yes, it's safe to replace [0;32] with a null in most practical contexts.

However, requiring the optimization in the spec doesn't feel right to me: this approach will break a natural property `forall a : Account . decodeAccount(encodeAccount(a)) == a`.

One benefit of requiring this optimization is that we get a canonical representation (at least until we introduce variable-length subaccounts with padded zeros).

But then, will we fail to explicitly parse textual representations that encode the default accounts? Or "optimize" them as well?

I'd keep the spec as simple as possible, allowing apps to pre-process subaccounts if they want to.

qti3e: That makes sense, but just to double-check for my understanding, the textual representation for an account with explicit subaccount zero is going to look different by the end user than when we do not have a subaccount at all?

README.md
MarioDfinity: @timohanke this is just a UI representation of the full accounts. It's likely to be just copied and sent around. I don't see the problem with this representation and it has nice properties.

MarioDfinity (approved):

README.md
roman-kashitsyn: > the textual representation for an account with explicit subaccount zero is going to look different by the end user than when we do not have a subaccount at all?

Yes, that was the plan.

One problem that comes to mind is potential user confusion when you send to an address and then see a different address in a block explorer because the tooling applied an optimization.

So having a unique encoding might be desirable. But then we need to reject unoptimized encodings in the decoder.

MarioDfinity (approved):

roman-kashitsyn:
@timohanke, @bogwar, I've updated the spec as discussed on the [forum](https://forum.dfinity.org/t/ledger-tokenization-working-group-update/15298/8) and included a reference implementation of encoder and decoder in Motoko.

README.md
```diff
@@ -221,37 +221,62 @@ The minting account is also the receiver of the fees burnt in regular transfers.
...
+3. A typo in the textual encoding invalidates it with a high probability.
 
 ### Encoding
 
 Applications SHOULD encode accounts as follows:
 
-  1. The encoding of accounts that do not have a subaccount, such as `{ owner = principal "aaaaa-aa"; subaccount = null }`, is the encoding of the owner principal.
-  2. The encoding of accounts with a non-empty subaccount is the encoding of the concatenation of the owner principal bytes, the subaccount bytes, and an extra byte `A0`<sub>`16`</sub> (160 = 128 + 32).
+  1. The encoding of the default account (the subaccount is null or a blob with 32 zeros) is the encoding of the owner principal.
+  2. The encoding of accounts with a non-default subaccount is the textual principal encoding of the concatenation of the owner principal bytes, the subaccount bytes with the leading zeros omitted, the length of the subaccount without the leading zeros (a single byte), and an extra byte `FF`<sub>16</sub>.
```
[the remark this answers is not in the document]
roman-kashitsyn: Done!

PR #95, opened 2023-02-26
Icrc4

skilesare:
Batch functions to complement ICRC-1.

### DIFF
--- a/standards/ICRC-4/readme.md
+++ b/standards/ICRC-4/readme.md
@@ -101,6 +101,7 @@ type TransferBatchError = variant {
 #### Postconditions
 
 * The accounts are updated with a new balance unless the pre_validate was requested and failed.
+* the results of each transaction are paired with the original request arguments. Ordering is not guaranteed.
 
 ### query icrc4_balance_of_batch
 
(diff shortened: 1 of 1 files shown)


skilesare:
Initial thoughts and more detailed reasoning:

> icrc4_transfer_batch needs to achieve atomicity.
> In batch transfers, either all succeed or all fail.

This is impossible to guarantee on the IC.  Future token canisters may be multicenter and involve one or many async calls.  ICRC1 makes no assumptions about token canister architecture and it would a complication to introduce this assumption.

> I don't think pre_validate and batch_fee need to be included in the standard, follow the rules of TransferArgs and icrc1_transfer, but here is an array, so TransferBatchArgs = vec TransferArgs would be better.

Because atomicity is not possible to guarantee, I'd argue the prevalidate provides the next best alternative to attempting to have as many transactions pass as possible.  I don't think it is optional for this kind of transaction.  the batch fee is optional and can be omitted, but provides a rational pathway to the reality that submitting 1000 batch transactions does not cost 1000x a single transaction.  It decreases the cost of moving money and increases the possible uptake of the token.

> There is no need to configure these limit parameters.

Given the reality of how the IC works, I'd say that it is highly advisable to have these limits defined by the service. Cycle limits may apply and ledgers may want to reject a certain amount of transactions so as to not block the canister.  They are optional so a ledger can choose to not provide them, but a production-level ledger expecting significant transactions my need to keep from being attacked, and defining the interface lets providers agree on how to check the ledger.

Issue #93, opened 2023-02-24
ICRC-2 Expiration Inconsistencies & long-term unused approval allowance build-up


levifeldman:
Scenario: 

- At time t, principal-A calls icrc2_approve(spender: principal-B, amount: 10, expiration: t+4). 
- At time t+2, principal-A calls icrc2_approve(spender: principal-B, amount: 20, expiration: t+9).
- At time t+3, icrc2_allowance(account: principal-A, spender: principal-B) -> { allowance: 30; expires_at: t+9; }.
- At time t+5, icrc2_allowance(account: principal-A, spender: principal-B) -> { allowance: 30; expires_at: t+9; }.
- At time t+6, principal-B calls icrc2_transfer_from(from: principal-A, to: principal-B, amount: 17);
- At time t+7, icrc2_allowance(account: principal-A, spender: principal-B) -> { allowance: 13; expires_at: t+9; }.

In this scenario, the first approval made at time t never got transferred (because principal-B didn't need it, or principal-B's auction/sale sold-out/finished). Most of the second approval was transferred at time t+6. At time t+7, principal-B still has the allowance of the first approval even after the first approval's expiration-time of t+4.

Since most unused approval allowances will not be reset because there is a fee to lower an allowance or to reset an allowance to 0 and because why make the extra call and because most will expect the ledger to expire an approval allowance at the set expiration time, the above scenarios can add up in the long term where spenders have huge allowances from built-up unused approvals because the expirations carried over to the latest approvals. 

Possible Solutions:

a: Remove the expiration option from an approval. This will remove the expiration inconsistencies but is still susceptible to spenders building up huge allowances from unused approvals. 

b: Honor each approval's expiration by separating the allowances of each approval. To transfer_from an account, a spender must tell the ledger the block-height/id of the approval.

changes: 
```candid

type TransferFromArgs = record {
    from : Account;
    to : Account;
    // new field
    approval_id: nat;
    amount : nat;
    fee : opt nat;
    memo : opt blob;
    created_at_time : opt nat64;
};

type AllowanceArgs = record {
    account : Account;
    spender : principal;
    approval_id: nat; // new field
};

```
In the ledger: 
```rust
type ApprovalId = BlockHeight;

static APPROVALS: Map<(Account,Spender,ApprovalId), Approval>;

struct Approval {
    remaining_allowance: u64,
    expires_at: u64,
}
```

Let's go through the scenario above with this solution in place. 
[...]

Issue #22, opened 2022-07-16
Insecure `approve` method


niftyorca:
In the approve branch, the `approve` function should not overwrite the allowance, but rather increment/decrement it. In the overwriting scenario, exploits are possible by frontrunning the new approval update tx.

Reference
https://swcregistry.io/docs/SWC-114

MarioDfinity:
Hey @niftyorca , the `approve` function won't be part of the ICRC-1 standard and we decided to move the discussion to an extension of the standard. I'll close this for now.

PR #101, opened 2023-03-14
ICRC-2 Separate approval allowances and honor the set expiration in an approval-request.

levifeldman:
Fixes: #93 

This change separates approvals made for the same Account-Spender pair into separate allowances each with their own expiration of the expiration set in the specific approval request.

This change lets the ledger honor the expiration time of a specific approval-allowance even if multiple overlapping approvals are made for the same Account-Spender pair. 

Before this change, when multiple overlapping approvals were made for the same Account-Spender pair, the ledger would have set the expiration of the spender's allowance to the expiration of the latest approval request, carrying over the allowance from the previous approval requests but not carrying over the expiration of the previous approval requests, therefore giving the spender the allowance from an approval-request that may have been expired. 

The solution of this change gives each approval a separate allowance & expiration even for the same Account-Spender pair. 

When a Spender calls `icrc2_transfer_from`, the ledger calculates the **total**-allowance for the Account-Spender pair by taking the **sum** of the current (unexpired) allowances for that Account-Spender pair. Then the ledger goes through the allowances one by one starting with the allowance expiring soonest and deducts the transfer amount + fee from the total-allowance. 

The `TransferFromArg` stays the same.

The `icrc2_allowances` method now returns the total-allowance for the Account-Spender pair and can show each specific approval allowance and expiration.

The `icrc2_cancel_approval` method can be used by an approver to cancel an existing approval. Any remaining allowance of that specific approval will be canceled. The method is called using the original approval block id as the parameter, and returns with the block-id of the approval-cancellation.

### DIFF
--- a/standards/ICRC-2/ICRC-2.did
+++ b/standards/ICRC-2/ICRC-2.did
@@ -6,7 +6,7 @@ type Account = record {
 type ApproveArgs = record {
     from_subaccount : opt blob;
     spender : principal;
-    amount : int;
+    amount : nat;
     expires_at : opt nat64;
     fee : opt nat;
     memo : opt blob;
@@ -45,9 +45,30 @@ type TransferFromError = variant {
     GenericError : record { error_code : nat; message : text };
 };
 
-type AllowanceArgs = record {
+type AllowancesArgs = record {
     account : Account;
     spender : principal;
+    start : opt nat;
+};
+
+type AllowancesData = record {
+    latest_approval_id : opt nat;
+    total_allowance : nat;
+    allowances : vec Allowance;
+};
+
+type Allowance = record {
+  approval_id: nat;
+  allowance : nat;
+  expires_at : opt nat64;
+};
+
+
+type CancelApprovalError = variant {
+    ApprovalNotFound;
+    CallerIsNotTheApprover;
+    ApprovalIsExpired;
+    ApprovalIsCanceled : record { cancellation: nat; };
 };
 
 service : {
@@ -55,5 +76,6 @@ service : {
 
     icrc2_approve : (ApproveArgs) -> (variant { Ok : nat; Err : ApproveError });
     icrc2_transfer_from : (TransferFromArgs) -> (variant { Ok : nat; Err : TransferFromError });
-    icrc2_allowance : (AllowanceArgs) -> (record { allowance : nat; expires_at : opt nat64 }) query;
+    icrc2_allowances : (AllowancesArgs) -> (AllowancesData) query;
+    icrc2_cancel_approval : (nat) -> (variant { Ok : nat; Err : CancelApprovalError });
 }
--- a/standards/ICRC-2/ICRC-2.did
+++ b/standards/ICRC-2/ICRC-2.did
@@ -67,8 +67,6 @@ type Allowance = record {
 type CancelApprovalError = variant {
     ApprovalNotFound;
     CallerIsNotTheApprover;
-    ApprovalIsExpired;
-    ApprovalIsCanceled : record { cancellation: nat; };
 };
 
 service : {
--- a/standards/ICRC-2/README.md
+++ b/standards/ICRC-2/README.md
@@ -10,7 +10,7 @@
 `ICRC-2` specifies a way for an account owner to delegate token transfers to a third party on the owner's behalf.
 
 The approve and transfer-from flow is a 2-step process.
-1. Account owner Alice entitles principal Bob to transfer up to X tokens from her account A by calling the `icrc2_approve` method on the ledger.
+1. Account owner Alice creates an approval for principal Bob to transfer up to X tokens from her account A by calling the `icrc2_approve` method on the ledger.
 2. Bob can transfer up to X tokens from account A to any account by calling the `icrc2_transfer_from` method on the ledger.
    The number of transfers Bob can initiate from account A is not limited as long as the total amount spent is below X.
 
@@ -46,16 +46,13 @@ type Account = record {
 
 ### icrc2_approve
 
-Entitles the `spender` to transfer an additional token `amount` on behalf of the caller from account `{ owner = caller; subaccount = from_subaccount }`.
-The number of transfers the `spender` can initiate from the caller's account is unlimited as long as the total amounts and fees of these transfers do not exceed the allowance.
+The caller creates an approval with an account `{ owner = caller; subaccount = from_subaccount }` with an amount for a `spender`.
+The ledger creates a new approval with the approval-transaction-block-id as the approval-id with the `amount` and `expires_at`.
+The ledger creates a new entry in the approvals-map: `Map<(Account,Spender,ApprovalId), Approval{expires_at: args.expires_at, available_allowance: args.amount}>`.
+When the expires_at field is set, the ledger cancels the approval-id at the expiration time and any remaining/unspent allowance of that approval allowance expires.
 The caller does not need to have the full token `amount` on the specified account for the approval to succeed, just enough tokens to pay the approval fee.
-The `spender`'s allowance for the account increases or decreases by the `amount` depending on the sign of the `amount` field. 
-If the `expires_at` field is not null, the ledger resets the approval expiration time to the specified value.
-
 The ledger SHOULD reject the request if the caller is the same principal as the spender (no self-approvals allowed).
 
-The ledger MAY cap the total allowance if it becomes too large (for example, larger than the total token supply).
-For example, if there are only 100 tokens, and the ledger receives two approvals for 60 tokens for the same `(account, principal)` pair, the ledger may cap the total allowance to 100.
 
 ```candid "Methods" +=
 icrc2_approve : (ApproveArgs) -> (variant { Ok : nat; Err : ApproveError });
@@ -65,7 +62,7 @@ icrc2_approve : (ApproveArgs) -> (variant { Ok : nat; Err : ApproveError });
 type ApproveArgs = record {
     from_subaccount : opt blob;
     spender : principal;
-    amount : int;
+    amount : nat;
     expires_at : opt nat64;
     fee : opt nat;
     memo : opt blob;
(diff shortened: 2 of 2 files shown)


iclighthouse:
The goal of the standard is to achieve interoperability and try to put only common interfaces inside the standard definition. I don't think it is appropriate to define the standard in terms of making a product.

If icrc2_approve is defined in the way described above, it increases the complexity of the implementation with limited improvement in security and has the negative effect of creating more data confusion.

Because Dapp uses icrc2_approve in a way that you can't force, for example, some dapps set both `expires_at` and `value` to large values to escape all the above restrictions, and users usually don't change it voluntarily.
A simple way to increase security is to allow the owner to query all approvals, so he can reset them. For example, add an `icrc2_approvals` method.

levifeldman:
> The goal of the standard is to achieve interoperability and try to put only common interfaces inside the standard definition. I don't think it is appropriate to define the standard in terms of making a product.

Interoperability requires consistent functionality. 

> A simple way to increase security is to allow the owner to query all approvals, so he can reset them. For example, add an icrc2_approvals method.

See the `icrc2_allowances` method and the `icrc2_cancel_approval` method in this PR. 

> Because Dapp uses icrc2_approve in a way that you can't force, for example, some dapps set both expires_at and value to large values to escape all the above restrictions, and users usually don't change it voluntarily.

@iclighthouse According to that thinking, it would be better if the ledger did not have to keep track of any allowance or expiration. We can make the ledger just handle delegations from an account to a spender so that the spender can spend unlimited amount from the approver’s account. Does that sound better to you?

iclighthouse:
`According to that thinking, it would be better if the ledger did not have to keep track of any allowance or expiration. We can make the ledger just handle delegations from an account to a spender so that the spender can spend unlimited amount from the approver’s account. Does that sound better to you?`
It doesn't say better. But did the altered solution solve the problem? No!
So, the effect before and after the change is almost the same, but the side effect is an increase in practical complexity. So why the change?

levifeldman:
> But did the altered solution solve the problem? No!
So, the effect before and after the change is almost the same, but the side effect is an increase in practical complexity. So why the change?

The solution in this PR does make sure that the allowance of an approval expires at the set expiration time in the original approval. Did you see the problem in #93?

> The icrc2_allowances method must specify the spender, but the user often cannot remember the spender address, and what the user needs is to get a list of all valid allowances before he can cancel.

I think its a good idea, we can make a method for a user to see all valid current allowances. Lets do it in a separate PR because this one is about solving #93 and because it is a good idea either way.

> The new scheme introduces such a large complexity without really solving the problem

This scheme solves the problem specified in #93.

> Also the data structure spender : principal will cause confusion of IDs, it is better to use spender : Account so that spender can specify its own subaccount when using icrc2_transfer_from, which provides room for business innovation for dapp developers. Developers only need to know that all IDs are Account.

I don't see a benefit of a spender specifying a subaccount in the icrc2_transfer_from method. Is there a sample use case where the spender specifying a subaccount can be beneficial?

iclighthouse:
> The solution in this PR does make sure that the allowance of an approval expires at the set expiration time of the original approval. Did you see the scenario in https://github.com/dfinity/ICRC-1/issues/93?

`approve` is usually used in a Dapp. The above statement assumes that the developer of the Dapp is a good guy and assumes that he will set the parameters reasonably. If the Dapp developer was a good guy, then the original solution would have no security issues and the interface would be simpler.
But security incidents are often the result of Dapp developers being cheaters, and they will set `approve` parameters that are conducive to stealing money. So the solution you described doesn't solve the problem at all, it only increases the complexity. Security is only possible in the hands of the user himself, who needs to manage all his approvals, which is safe enough and does not require the Dapp developer to set more parameters.
You are using a wrong approach to solve an impossible problem.

> I don't see a benefit of a spender specifying a subaccount in the icrc2_transfer_from method. Is there a sample use case where the spender specifying a subaccount can be beneficial?

We are DEX developers. There are multiple businesses in a Dex, each with its own dedicated subaccount. for security, I want the user to approve not to a canister-id, but to a business-specific subaccount. so that the allowance of business 1, is not spent by business 2 (since the user did not approve business 2 to spend it). Note: Spender is often a Canister, not a self-authenticating account.
I would also say that standards tend to prioritize a path that responds to more possibilities with a low-complexity approach so that the market has enough innovation to emerge.

levifeldman:
>> The solution in this PR does make sure that the allowance of an approval expires at the set expiration time of the original approval. Did you see the scenario in https://github.com/dfinity/ICRC-1/issues/93?

> `approve` is usually used in a Dapp. The above statement assumes that the developer of the Dapp is a good guy and assumes that he will set the parameters reasonably.

The statement does not make any assumptions about the developer of the dapp. The statement states that the solution in the PR makes sure that the allowance of an approval expires at the set expiration time of the original approval request. 
If the dapp developer is in control of the users account then the dapp can use `icrc1_transfer`. What is the point of `icrc2` then?

iclighthouse:
When icrc1_transfer() is executed, the wallet prompts the user who to send it to, how much to send, and then signs it to be sure. At this point the user will perceive the danger. It would be much easier to find a reason for the user to approve.

> Of course, you are talking about the agentjs interaction, which cannot be applied to asset-based applications, a problem I have communicated to Ben. This is very dangerous, this can't be done for Defi applications because it must assume that the Dapp developer is trustworthy. This is another topic where IC needs to rethink the interaction style to fit the Defi application scenario.

You can check the history of `approve` security incidents in ERC20. There are two general categories:
1) Phishing to get the user's allowance and then transferring assets.
2) Allowance is set in a normal business requirement, and after a period of time, the dapp developer decides to do something evil for some reason and then steals the funds.

So the security remains in the hands of the user:
1) Users should not give allowance until they make sure the application code is trusted.
2, After finishing the operation, go to the wallet as soon as possible to check all the allowances and cancel the unnecessary ones.

iclighthouse:
There is another solution with even fewer changes that can achieve the same effect. `expires_at` was already in the original scheme, now it's just a matter of changing `spender` to type `Account` and then `icrc2_transfer_from` adding the `spender_subaccount` parameter.

Different allowances can be achieved by specifying different subaccounts, and the subaccount is fully specified by Dapp with predictability, unlike the unpredictability of approval-id.

levifeldman:
> unlike the unpredictability of approval-id

The solution in this PR does **not** require the spender to know the approval-id.

[Here](https://github.com/levifeldman/ICRC-1/tree/icrc2-separate-approval-balances/standards/ICRC-2#icrc2_transfer_from) is how the ledger handles the `icrc2_transfer_from` function.

iclighthouse:
A. "Separate approval allowances"
B. "Define the data type of the `spender` as Account (principal + subaccount)"

Please provide a full evaluation of the differences and advantages of these two options.
If you feel that A is better, I would be happy to hear the reasons that convince me. If, for some considered factor, you have settled on option A, then please just let me know and I will not comment further here, I do not mean to discourage option A.

It's just that I think option B is better because:
1. the two solutions can achieve almost the same effect, the difference is that solution A can combine allowances when using them, solution B separates both spenders and allowancesd.
2. The complexity of implementing Tokens in solution B is lower and simpler.
3. Solution B makes the development of dapp less complex

iclighthouse:
Based on the experience of developing DEX, most defi projects do not have a need for #93, and it is a very effective measure to overwrite all historical settings when setting up an allowance. This is also a common practice among most defi developers, and changing such habits carries risks in itself.

levifeldman:
> A. "Separate approval allowances"
B. "Define the data type of the spender as Account (principal + subaccount)"
Please provide a full evaluation of the differences and advantages of these two options.

These two options do not contradict each other. adding a subaccount to the spender can be done whether or not the approval allowances are separate.

roman-kashitsyn:
@iclighthouse 

In the ICRC-2 design, we decided to make approvals _additive_ (an idea originally proposed in #22 way before we started working on ICRC-2). The motivation for that is to make _concurrent_ approvals work nicely together.
For example, if we use the ERC-20 semantics and a pair of canister trades at high speed, the following scenario is possible:
1. A approves 10 ICP to B for transaction X.
2. Ledger sees A's requests and sets (A,B) approval to 10 ICP.
3. B attempts to execute a transfer of 10 ICP within the context of transaction X.
4. A approves 15 ICP to B for transaction Y.
5. Ledger sees A's request and sets (A,B) approval to 15.
6. Ledger sees B's request and transfers 10 ICP from A's account, setting (A,B) approval to 5 ICP.
7. B attempts to execute a transfer of 15 ICP within the context of transaction Y.
8. Ledger sees B's request and rejects the transfer because of an insufficient allowance.

With additive approvals, the Ledger will process both B's transfers correctly.

Another improvement upon ERC-20 is the expiration dates for approvals. I'd argue that _overwriting_ expiration dates contradicts the design choice of independent approvals. The additive semantics meant to make approvals independent.

I see four options we could implement:

1. Use ERC-20 as-is without extensions.
2. Start with ERC-20 and implement additive semantics of approvals but do not introduce expiration times.
3. Start with ERC-20 and implement expiration times, keeping the original semantics for amounts. I.e., each call to `approve` overwrites the results of the previous call.
4. Introduce additive approval semantics _and_ individual expiration times for approvals. That way, we have the most intuitive design when each approval acts independently on both dimensions and does not interfere much with other approvals.

I'm OK with any of the four outcomes above, but I'd argue against mixing additive approvals with overwriting expiration dates. I, as a Dapp developer myself, find it very confusing. I expect many other developers to have the same issue.

iclighthouse:
Fine-grained control of allowances requires dapp developers to call icrc2_approve and icrc2_allowances more frequently, which can compromise on the user experience. So it has become a usage convention for many developers to set large numbers on the first call to icrc2_approve and then go back and recheck allowances when icrc2_transfer_from reports an error.
ETH's web3js calls are also asynchronous, the asynchronous calls don't cause conflicts and you can cache allowances and amounts spent on the front end.

timohanke:
My preferred option is:
- implement two functions `set` and `add` where `set` is absolute and `add` is incremental (and returns the new value)
- allow to set an expiration time but a global one, not individual ones on a per-approval basis, i.e. a new expiration time overwrites the old one

Rationale: there are two different reasons to have expiration times:
1) You want to try out a service, give it some approval, and when you stop using it you might forget to delete the approval, so an expiration protects you against forgetting it.
2) You want to schedule approvals for the future, basically define a function over time, that could go up or down, to manage things like recurring approvals (absolute or incremental), approvals for transaction contexts, etc.

In my point of view 2) is simply bottom-less. There will be arbitrarily complex demands. Hence, just do 1). 

Setting the an absolute value and incrementing/decrementing by a delta (same as atomic read and set) will both be needed in certain contexts. In particular set to 0 will be needed.

roman-kashitsyn:
Thanks a lot for the input!

The working group discussed the issue on 2023-04-18 and agreed to stay closer to the ERC-20 semantics to keep the implementation simple and avoid confusing developers from the Ethereum ecosystem.
I captured the outcome of the discussion in the following PR: https://github.com/dfinity/ICRC-1/pull/109

levifeldman:
Thanks for the good conversation everyone! As Roman mentions, PR-https://github.com/dfinity/ICRC-1/pull/109 fixes the issue in https://github.com/dfinity/ICRC-1/issues/93 and removes the need for this PR-https://github.com/dfinity/ICRC-1/pull/101.

@iclighthouse Thanks for bringing up the possibility of using subaccounts for the spender! Now that PR-https://github.com/dfinity/ICRC-1/pull/109 uses overwrite semantics instead of additive semantics, using subaccounts for the spender is a great way of separating approvals for the same spender.

levifeldman:
https://github.com/dfinity/ICRC-1/pull/109 is merged into main.

PR #109, opened 2023-04-19
feat: use ERC-20 semantics for approvals

roman-kashitsyn:
This change incorporates the feedback from the Working Group session on 2023-04-18:

  1. We change the semantics of approvals to be closer to ERC-20. A call to icrc2_approve overwrites the previous allowance and the timestamp.

   2. Allowances target accounts, not principals. This change will enable fine-grained control in DApps and significantly simplify implementation for ledgers that rely on AccountIdentifiers internally, such as the ICP ledger.

   3. For applications that need concurrent approvals, we add optional "compare and swap" semantics of approvals. If the caller specifies the optional `expected_allowance` field, the ledger will return an error if the allowance changes before the ledger accepts the `approve` call.

TODO: update the reference implementation once we agree on the semantics.

### DIFF
--- a/standards/ICRC-2/README.md
+++ b/standards/ICRC-2/README.md
@@ -197,7 +197,7 @@ icrc1_supported_standards : () -> (vec record { name : text; url : text }) query
 
 1. Canister C wants to transfer 100 tokens on an `ICRC-2` ledger from Alice's account to Bob's account.
 2. Alice previously approved canister C to transfer tokens on her behalf by calling `icrc2_approve` with `spender` set to the canister's default account (`{ owner = C; subaccount = null }`) and `amount` set to the token amount she wants to allow (100) plus the transfer fee.
-3. During some update call, the canister can now call `icrc2_transfer_from` with `from` set to Alice's account, `to` set to Bob's account, and `amount` set to the token amount she wants to transfer (100 plus the transfer fee).
+3. During some update call, the canister can now call `icrc2_transfer_from` with `from` set to Alice's account, `to` set to Bob's account, and `amount` set to the token amount she wants to transfer (100).
 4. Once the call completes successfully, Bob has 100 extra tokens on his account, and Alice has 100 (plus the fee) tokens less in her account.
 
 ### Alice removes her allowance for canister C
--- a/standards/ICRC-2/README.md
+++ b/standards/ICRC-2/README.md
@@ -53,7 +53,7 @@ The number of transfers the `spender` can initiate from the caller's account is 
 The caller does not need to have the full token `amount` on the specified account for the approval to succeed, just enough tokens to pay the approval fee.
 The call resets the allowance and the expiration date for the `spender` account to the given values.
 
-The ledger SHOULD reject the call if the spender account is equal to the source account (`{ owner = caller; subaccount = from_subaccount } == spender`).
+The ledger SHOULD reject the call if the spender account owner is equal to the source account owner.
 
 If the `expected_allowance` field is set, the ledger MUST ensure that the current allowance for the `spender` from the caller's account is equal to the given value and return the `AllowanceChanged` error otherwise.
 
--- a/standards/ICRC-2/README.md
+++ b/standards/ICRC-2/README.md
@@ -10,9 +10,11 @@
 `ICRC-2` specifies a way for an account owner to delegate token transfers to a third party on the owner's behalf.
 
 The approve and transfer-from flow is a 2-step process.
-1. Account owner Alice entitles principal Bob to transfer up to X tokens from her account A by calling the `icrc2_approve` method on the ledger.
-2. Bob can transfer up to X tokens from account A to any account by calling the `icrc2_transfer_from` method on the ledger.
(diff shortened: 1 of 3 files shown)


MarioDfinity (approved):

standards/ICRC-2/README.md
```diff
@@ -46,16 +48,17 @@ type Account = record {
...
 
-Entitles the `spender` to transfer an additional token `amount` on behalf of the caller from account `{ owner = caller; subaccount = from_subaccount }`.
+This method entitles the `spender` to transfer token `amount` on behalf of the caller from account `{ owner = caller; subaccount = from_subaccount }`.
 The number of transfers the `spender` can initiate from the caller's account is unlimited as long as the total amounts and fees of these transfers do not exceed the allowance.
 The caller does not need to have the full token `amount` on the specified account for the approval to succeed, just enough tokens to pay the approval fee.
-The `spender`'s allowance for the account increases or decreases by the `amount` depending on the sign of the `amount` field. 
-If the `expires_at` field is not null, the ledger resets the approval expiration time to the specified value.
+The call resets the allowance and the expiration date for the `spender` account to the given values.
+
+The ledger SHOULD reject the call if the spender account is equal to the source account (`{ owner = caller; subaccount = from_subaccount } == spender`).
```
levifeldman, line 56:
```suggestion
The ledger SHOULD reject the call if the spender account owner is equal to the source account owner.
```

standards/ICRC-2/README.md
```diff
@@ -175,27 +185,34 @@ icrc1_supported_standards : () -> (vec record { name : text; url : text }) query
...
-2. Alice previously approved the canister to transfer tokens on her behalf by calling `icrc2_approve` with `spender` set to the canister's principal and `amount` set to the token amount she wants to allow (100) plus the transfer fee.
-3. During some update call, the canister can now call `icrc2_transfer_from` with `from` set to Alice's account, `to` set to Bob's account, and `amount` set to the token amount she wants to transfer (100).
-4. Depending on the result of the call, Bob now has 100 tokens in his account, and Alice has 100 tokens less in her account.
+1. Canister C wants to transfer 100 tokens on an `ICRC-2` ledger from Alice's account to Bob's account.
+2. Alice previously approved canister C to transfer tokens on her behalf by calling `icrc2_approve` with `spender` set to the canister's default account (`{ owner = C; subaccount = null }`) and `amount` set to the token amount she wants to allow (100) plus the transfer fee.
+3. During some update call, the canister can now call `icrc2_transfer_from` with `from` set to Alice's account, `to` set to Bob's account, and `amount` set to the token amount she wants to transfer (100 plus the transfer fee).
```
levifeldman, line 200:
```suggestion
3. During some update call, the canister can now call `icrc2_transfer_from` with `from` set to Alice's account, `to` set to Bob's account, and `amount` set to the token amount she wants to transfer (100).
```
The transfer fee is put into the `fee` field and not added in the `amount` field when Canister c is transfer_from-ing from alice's account to bob's account.

levifeldman (changes requested):

levifeldman:
Great PR! 

This PR is a solution for the issue in https://github.com/dfinity/ICRC-1/issues/93.

standards/ICRC-2/README.md
roman-kashitsyn: Initially, I restricted the rejection to only the same account because it might simplify the implementation of some ledgers (the ICP ledger, for example).

In a ledger that works with AccountIdentifiers, it's hard to check whether an account belongs to a particular principal. 

It's still possible to implement the check you propose; we have to do it much earlier before the conversion to AccountIdentifier happens. I'm OK either way.

levifeldman: I see how it can simplify the implementation of ledgers that work with account-identifiers if the implementation first converts icrc accounts to account-identifiers before the main method logic however since the icrc2 standard works with visible account owners and the icrc2_approve method request contains the approver/caller account owner-principal and the spender account owner-principal, I rather the standard make the check. If a caller of icrc2_approve can approve to him/her self, why not let the approver-account be the same as the spender account?

standards/ICRC-2/README.md
```diff
@@ -89,19 +96,17 @@ type ApproveError = variant {
...
 
 * The caller has enough fees on the `{ owner = caller; subaccount = from_subaccount }` account to pay the approval fee.
-* The caller is different from the `spender`.
-* If the `expires_at` field is set, it's creater than the current ledger time.
+* If the `expires_at` field is set, it's greater than the current ledger time.
+* If the `expected_allowance` field is set, it's equal to the current allowance for the `spender`.
 
 #### Postconditions
 
 * The `spender`'s allowance for the `{ owner = caller; subaccount = from_subaccount }` increases by the `amount` (or decreases if the `amount` is negative).
```
roman-kashitsyn, line 104:
```suggestion
* The `spender`'s allowance for the `{ owner = caller; subaccount = from_subaccount }` is equal to the given `amount`.
```

PR #119, opened 2023-06-06
fix/icrc2_approve-method

h1teshtr1path1:
I can see assertion in 2nd last line in icrc2_approve method will always throw err, am I missing something?

### DIFF
--- a/ref/ICRC1.mo
+++ b/ref/ICRC1.mo
@@ -6,6 +6,7 @@ import Option "mo:base/Option";
 import Error "mo:base/Error";
 import Time "mo:base/Time";
 import Int "mo:base/Int";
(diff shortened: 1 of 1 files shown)


roman-kashitsyn:
The assertions states that the approver balance goes down by `effectiveFee` once the ledger records the Approve transaction. Do you observe this assertion failing?

h1teshtr1path1:
Yes, assertion was failing.
But I think It's because only init.minter_account is allowed to call icrc2_approve function and give allowance to others, and since init.minting_account → tx_fee is always zero, that why this assertion won't fail for init.minting account. Am I right?

Because if tx_fee is not zero, it will fail in all cases, right?

PR #127, opened 2023-07-11
feat(FI-813)  changed LedgerEnv to be a trait

NikolasHai:
This MR introduces the following changes:

- Change LedgerEnv to be a trait
- Create the struct LedgerEnvReplica which connects to a running replica
- Implement the trait LedgerEnv for the LedgerEnvReplica


test/env/lib.rs
```diff
@@ -147,8 +148,27 @@ impl Transfer {
...
+    async fn transfer(&self, arg: Transfer) -> anyhow::Result<Result<Nat, TransferError>>;
+    fn principal(&self) -> Principal;
+    async fn balance_of(&self, account: impl Into<Account>) -> anyhow::Result<Nat>;
+    async fn supported_standards(&self) -> anyhow::Result<Vec<SupportedStandard>>;
+    async fn metadata(&self) -> anyhow::Result<Vec<(String, Value)>>;
+    async fn minting_account(&self) -> anyhow::Result<Option<Account>>;
+    async fn token_name(&self) -> anyhow::Result<String>;
+    async fn token_symbol(&self) -> anyhow::Result<String>;
+    async fn token_decimals(&self) -> anyhow::Result<u8>;
+    async fn transfer_fee(&self) -> anyhow::Result<Nat>;
```
roman-kashitsyn, line 163: It's usually beneficial to keep traits as small as possible.

Note that the `query` endpoint is enough to implement many other methods.
I suggest we keep only `query` and `update` (it's not in the interface yet, but could be) methods in the interface and implement everything else as either free functions or extension traits.

NikolasHai: I split LedgerEnv into two traits (LedgerEnv,LedgerEndpoint):
I would argue that `principal()` and `fork()` also need to be in the LedgerEnv trait as they are used in the tests multiple times and in the tests we only specify the traits not the struct implementation. Free-floating functions would only be possible for the `new(...)` operator. 

I also added the `update(...)` function to LedgerEnv

test/env/lib.rs
```diff
@@ -147,8 +148,35 @@ impl Transfer {
...
+pub trait LedgerEndpoint: LedgerEnv {
+    async fn transfer(&self, arg: Transfer) -> anyhow::Result<Result<Nat, TransferError>>;
+    async fn balance_of(&self, account: impl Into<Account>) -> anyhow::Result<Nat>;
+    async fn supported_standards(&self) -> anyhow::Result<Vec<SupportedStandard>>;
+    async fn metadata(&self) -> anyhow::Result<Vec<(String, Value)>>;
+    async fn minting_account(&self) -> anyhow::Result<Option<Account>>;
+    async fn token_name(&self) -> anyhow::Result<String>;
+    async fn token_symbol(&self) -> anyhow::Result<String>;
+    async fn token_decimals(&self) -> anyhow::Result<u8>;
+    async fn transfer_fee(&self) -> anyhow::Result<Nat>;
```
roman-kashitsyn, line 175: There should be only one "right" implementation for all of these methods in terms of LedgerEnv::query and LedgerEnv::update, so let's provide the default implementation for all methods.
This way, we won't need to repeat the exact implementation for all instances of LedgerEnv.

test/env/lib.rs
```diff
@@ -147,8 +148,35 @@ impl Transfer {
...
+        Input: ArgumentEncoder + std::fmt::Debug,
+        Output: for<'a> ArgumentDecoder<'a>;
+    async fn update<Input, Output>(&self, method: &str, input: Input) -> anyhow::Result<Output>
+    where
+        Input: ArgumentEncoder + std::fmt::Debug,
+        Output: for<'a> ArgumentDecoder<'a>;
+}
+
+#[async_trait(?Send)]
+pub trait LedgerEndpoint: LedgerEnv {
```
roman-kashitsyn, line 166:
```suggestion
pub trait Icrc1Endpoints: LedgerEnv {
```

MarioDfinity, line 166: I'm not sure this should be a trait. The logic of all ICRC-1 and ICRC-2 methods should be the same except for the `LedgerEnv`. I would expect functions like `transfer` to not be part of a trait and instead take in input the `LedgerEnv`.

roman-kashitsyn: That also works for me; I don't have strong preferences.

test/env/lib.rs
```diff
@@ -269,3 +264,61 @@ impl LedgerEnv {
...
+    ledger.query("icrc1_symbol", ()).await.map(|(t,)| t)
+}
+
+pub async fn token_decimals(ledger: &impl Ledger) -> anyhow::Result<u8> {
+    ledger.query("icrc1_decimals", ()).await.map(|(t,)| t)
+}
+
+pub async fn transfer_fee(ledger: &impl Ledger) -> anyhow::Result<Nat> {
+    ledger.query("icrc1_fee", ()).await.map(|(t,)| t)
+}
```
roman-kashitsyn, line 324: nit: I'd package these functions into an `icrc1` module.
Something like

```rust
pub mod icrc1 {
    pub async fn transfer(...) 
    pub async fn balance_of(...)
}
```

test/env/lib.rs
```diff
@@ -147,8 +148,25 @@ impl Transfer {
...
+    async fn update<Input, Output>(&self, method: &str, input: Input) -> anyhow::Result<Output>
+    where
+        Input: ArgumentEncoder + std::fmt::Debug,
+        Output: for<'a> ArgumentDecoder<'a>;
+}
+
+pub trait LedgerEnv {
+    fn fork(&self) -> Self;
+    fn principal(&self) -> Principal;
+}
```
roman-kashitsyn, line 166: I'd merge these traits back for now because we rarely want to use them separately.

roman-kashitsyn (approved):
LGTM, thanks!

PR #128, opened 2023-07-11
ICRC-3

MarioDfinity:


standards/ICRC-3/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+This hash function should be used by Ledgers to calculate the hash of the parent of a transaction and by clients to verify the downloaded transaction log.
+
+The hash function works is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
+- the hash of a `Blob` is the hash of the bytes themselves
+- the hash of a `Text` is the hash of the bytes representing the text
+- the hash of a `Nat` is the [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128) encoding of the number
+- the hash of an `Int` is the [`sleb128`](https://en.wikipedia.org/wiki/LEB128#Signed_LEB128) encoding of the number
+- the hash of an `Array` is the hash of the concatenation of the hashes of all the elements of the array
+- the hash of a `Map` is the hash of the concatenation of all the hashed items of the map sorted. An hashed item is the tuple composed by the hash of the key and the hash of the value.
```
bogwar, line 60:
```suggestion
- the hash of a `Map` is the hash of the concatenation of all the hashed items of the map sorted. A hashed item is the tuple composed by the hash of the key and the hash of the value.
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Standards Transactions
```
bogwar, line 62:
```suggestion
## Interaction with other standards
```

blutooth:
Really important to be able to query txs by deterministic tx hashes. On ETH it is possible to do this, I hope the IC log contains this functionality, which is crucial for finding transactions.

sea-snake:
What would be the recommended approach for getting transactions from/to a specific account without fetching the transaction history and filtering?

Right now most wallets use `rosetta-api.internetcomputer.org` to get a list of user transactions, but this isn't on chain and won't work for other icrc1 tokens.

Is this something that will be handled in another ICRC standard?

MarioDfinity:
@sea-snake ICRC-3 is concerned with the transaction log only. You can write a canister that indexes transactions for wallets/transaction explorers consumption based on the ICRC-3 API. For instance, we implemented an [index canister](https://github.com/dfinity/ic/blob/629f7fc8e82ff52e6373a528dcd29f9066563ad4/rs/rosetta-api/icrc1/index-ng/index-ng.did) for our [ICRC Ledger](https://github.com/dfinity/ic/tree/629f7fc8e82ff52e6373a528dcd29f9066563ad4/rs/rosetta-api/icrc1/ledger) that allows the NNS Dapp to query an Account transactions via [`get_account_transactions`](https://github.com/dfinity/ic/blob/629f7fc8e82ff52e6373a528dcd29f9066563ad4/rs/rosetta-api/icrc1/index-ng/index-ng.did#L111). We are doing the same for the ICP Ledger.

We should discuss about an indexing standard in the L&T WG.

MarioDfinity:
@blutooth ICRC-3 doesn't provide a way to query transactions by transaction hash. This is a good topic for the WG.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Servers must serve the transaction log as list of `Value` where each `Value` represent a single transaction in the transaction log.
+
+## Value Hash
+
+`ICRC-3` specifies a standard hash function over `Value`.
+
+This hash function should be used by Ledgers to calculate the hash of the parent of a transaction and by clients to verify the downloaded transaction log.
+
+The hash function works is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
```
THLO, line 54:
```suggestion
The hash function is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` must define the list of transactions types they define and/or extend together with their `Value` schema and the function that converts a [`Value`](#value) to that type. Transaction types are well-typed records that are easy to consume by clients.
+
+`Value`s representing transactions must have the following schema properties:
+1. they must be of type `Map`
+1. they must have a field `tx: Map` which describes the transaction
+1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. "ICRC1_Burn" for `ICRC1_Burn` transactions
+1. all transactions with index >0 must have a top-level field called `phash: Blob` which contains the [hash](#value-hash) of the previous transactions.
```
THLO, line 70: Shouldn't this be a numbered list (1., 2., 3., ...)?

THLO, line 70:
```suggestion
1. they must be of type `Map`
1. they must have a field `tx: Map` which describes the transaction
1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. "ICRC1_Burn" for `ICRC1_Burn` transactions
1. all transactions with `index > 0` must have a top-level field called `phash: Blob` which contains the [hash](#value-hash) of the previous transactions. 
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` must define the list of transactions types they define and/or extend together with their `Value` schema and the function that converts a [`Value`](#value) to that type. Transaction types are well-typed records that are easy to consume by clients.
+
+`Value`s representing transactions must have the following schema properties:
+1. they must be of type `Map`
+1. they must have a field `tx: Map` which describes the transaction
+1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. "ICRC1_Burn" for `ICRC1_Burn` transactions
+1. all transactions with `index > 0` must have a top-level field called `phash: Blob` which contains the [hash](#value-hash) of the previous transactions.
```
bogwar, line 70:
```suggestion
1. each transaction with `index > 0` must have a top-level field called `phash: Blob` which contains the [hash](#value-hash) of the previous transaction. 
```

bogwar: I think we also need to specify what is `phash` for the first transaction since this is how one determines that it's the beginning of the chain. Either it's None, or a value with no hash preimage (maybe the all 0 blob).

hpeebles:
I'd imagine that many apps will store transactions for each user and want to pull in newer transactions when the user loads their transactions page. To do this they'd ideally be able to pass in the latest transaction Id they have stored for the user and request newer transactions than that.
This would be solved by ICRC-3 if we could optionally choose the direction of iteration.
Is this something you have considered adding?
An additional benefit of this is that if you are syncing transactions from a canister you avoid having to make a replicated query to first work out the latest block index.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+## Value Hash
+
+`ICRC-3` specifies a standard hash function over `Value`.
+
+This hash function should be used by Ledgers to calculate the hash of the parent of a transaction and by clients to verify the downloaded transaction log.
+
+The hash function is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
+- the hash of a `Blob` is the hash of the bytes themselves
+- the hash of a `Text` is the hash of the bytes representing the text
+- the hash of a `Nat` is the [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128) encoding of the number
```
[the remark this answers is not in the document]
bogwar: I think it should be the hash. Otherwise there's no bound on the hash of a Value and that looks undesirable and a somewhat inconsistent. Having said that, we'd never have to hash a raw Nat or Int; these would always be part of some outer structure which would be compressed.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+2. otherwise the `last_block_hash` in the tip certificate.
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` must define the list of transactions types they define and/or extend together with their `Value` schema and the function that converts a [`Value`](#value) to that type. Transaction types are well-typed records that are easy to consume by clients.
+
+`Value`s representing blocks must have the following schema properties:
+1. they must be of type `Map`
+1. they must have a field `tx: Map` which describes the transaction inside the block
+1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. "ICRC1_Burn" for `ICRC1_Burn` transactions
```
roman-kashitsyn, line 84:
```suggestion
1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. `"burn"` for `ICRC1_Burn` transactions
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+In other words the schema must be:
+
+```
+type ICRC1_Block = {
+    // The hash of the parent Transaction
+    "phash" : Blob?;
+
+    // The timestamp of when the block was
+    // added to the Ledger
+    "ts": u64,
```
roman-kashitsyn, line 96: nit: we use commas and semicolons inconsistenly in our little schema language. We should make our mind and use only one separator.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+
+#### ICRC2_Approve Schema
+
+```
+type ICRC2_Approve = ICRC1_Common and {
+  "op": "approve";
+  "from": Account;
+  "spender": Account;
+  "expected_allowance": u64?;
```
roman-kashitsyn, line 245:
```suggestion
  "expected_allowance": Nat?;
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Value
+
+The [candid](https://github.com/dfinity/candid) format supports sharing information even when the client and the server involved do not have the same schema (see the [Upgrading and subtyping](https://github.com/dfinity/candid/blob/master/spec/Candid.md#upgrading-and-subtyping) section of the candid spec). While this mechanism allows to evolve services and clients
+independently without breaking them, it also means that a client may not receive all the information that the server is sending, e.g. in case the client schema lacks some fields that the server schema has.
+
+This loss of information is not an option for `ICRC-3`. The client must receive the same exact data the server sent. For this reason, `ICRC-3` introduces the `Value` type which never changes:
+
+```
+type Value = variant {
```
roman-kashitsyn, line 39: I believe our variant also had `Nat64 : nat64`.

standards/ICRC-3/README.md
roman-kashitsyn, line 96:
```suggestion
    "ts": Nat64,
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+  "fee" : Nat?;
+
+  // The memo added by the user
+  "memo" : Blob?;
+
+  // The time at which the transaction
+  // was created by the user. When set,
+  // the Ledger must deduplicate the
+  // transaction
+  "ts" : u64?;
```
roman-kashitsyn, line 203:
```suggestion
  "ts" : Nat64?;
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### ICRC2_Approve Schema
+
+```
+type ICRC2_Approve = ICRC1_Common and {
+  "op": "approve";
+  "from": Account;
+  "spender": Account;
+  "expected_allowance": u64?;
+  "expires_at": u64?;
```
roman-kashitsyn, line 246:
```suggestion
  "expires_at": Nat64?;
```

roman-kashitsyn (commented):
One minor remark: I suspect many people might get confused about the schema definition language and how it's different from the Candid definitions. We might want to elaborate on the difference between these two languages and move all the schemas in a separate section so the reader doesn't have to switch back and forth between Candid and the Value Schema language.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+type Value = variant {
+    Blob : blob;
+    Text : text;
+    Nat : nat; // do we need this or can we just use Int?
+    Int : int;
+    Array : vec Value;
+    Map : vec record { text; Value };
+};
+
+type GetBlocksArgs = vec record { start : nat; length : nat };
```
seniorjoinu, line 132: How do I read _the latest_ block?
This is crucial for any block explorer.

roman-kashitsyn: `icrc3_get_tip_certificate` returns the latest block number together with its hash (both values are certified).

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+    Blob : blob;
+    Text : text;
+    Nat : nat; // do we need this or can we just use Int?
+    Int : int;
+    Array : vec Value;
+    Map : vec record { text; Value };
+};
+```
+
+Servers must serve the block log as list of `Value` where each `Value` represent a single block in the block log.
```
bogwar, line 49:
```suggestion
Servers must serve the block log as a list of `Value` where each `Value` represent a single block in the block log.
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+## Value Hash
+
+`ICRC-3` specifies a standard hash function over `Value`.
+
+This hash function should be used by Ledgers to calculate the hash of the parent of a block and by clients to verify the downloaded block log.
+
+The hash function is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
+- the hash of a `Blob` is the hash of the bytes themselves
+- the hash of a `Text` is the hash of the bytes representing the text
+- the hash of a `Nat` is the [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128) encoding of the number
```
bogwar, line 60:
```suggestion
- the hash of a `Nat` is the hash of the [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128) encoding of the number
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+- the hash of an `Int` is the [`sleb128`](https://en.wikipedia.org/wiki/LEB128#Signed_LEB128) encoding of the number
```
bogwar, line 61:
```suggestion
- the hash of an `Int` is the hash of the [`sleb128`](https://en.wikipedia.org/wiki/LEB128#Signed_LEB128) encoding of the number
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+- the hash of an `Array` is the hash of the concatenation of the hashes of all the elements of the array
+- the hash of a `Map` is the hash of the concatenation of all the hashed items of the map sorted lexicographically. A hashed item is the tuple composed by the hash of the key and the hash of the value.
+
+Pseudocode for representation independent hashing of Value, together with test vectors to check compliance with the specification can be found [`here`](HASHINGVALUES.md). 
+
+## Blocks Verification
+
+The Ledger MUST certify the last block (tip) recorded. The Ledger MUST allow to download the certificate via the `icrc3_get_tip_certificate` endpoint. The certificate follows the [IC Specification for Certificates](https://internetcomputer.org/docs/current/references/ic-interface-spec#certification). The certificate contains a tree with the certified data and the signature. The tree MUST contain two labelled values (leafs):
```
bogwar, line 69:
```suggestion
The Ledger MUST certify the last block (tip) recorded. The Ledger MUST allow to download the certificate via the `icrc3_get_tip_certificate` endpoint. The certificate follows the [IC Specification for Certificates](https://internetcomputer.org/docs/current/references/ic-interface-spec#certification). The certificate is comprised of a tree containing the certified data and the signature. The tree MUST contain two labelled values (leafs):
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+1. `last_block_index`: the index of the last block in the chain. The values must be expressed as big-endian
+2. `last_block_hash`: the hash of the last block in the chain
+
```
bogwar, line 72: Should how to download the ledger be part of the standard?  What if a client wants to download the ledger first and only then check the chaining?

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+Clients SHOULD download the tip certificate first and then download the block backward starting from `last_block_index` and validate the blocks in the process.
+
+Validation of block `i` is done by checking the block hash against
+1. if `i + 1 < len(chain)` then the parent hash `phash` of the block `i+1`
+2. otherwise the `last_block_hash` in the tip certificate.
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` must define the list of transactions types they define and/or extend together with their `Value` schema and the function that converts a [`Value`](#value) to that type. Transaction types are well-typed records that are easy to consume by clients.
```
bogwar, line 81: "and/or" is a bit strange. Should be "or" only, or not even talk about "extend".

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+`Value`s representing blocks must have the following schema properties:
+1. they must be of type `Map`
+1. they must have a field `tx: Map` which describes the transaction inside the block
+1. they must have a field `op: Text` inside `tx` which describes the type of transactions, e.g. "ICRC1_Burn" for `ICRC1_Burn` transactions
+1. all blocks with `index > 0` must have a top-level field called `phash: Blob` which contains the [hash](#value-hash) of the previous block.
+
+In other words the schema must be:
```
bogwar, line 89: Is this "in other words" or is this an example of how ICRC1_Blocks look like?

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+   └──────────────────────┘          └──────────────────────┘
+
+```
+
+## Value
+
+The [candid](https://github.com/dfinity/candid) format supports sharing information even when the client and the server involved do not have the same schema (see the [Upgrading and subtyping](https://github.com/dfinity/candid/blob/master/spec/Candid.md#upgrading-and-subtyping) section of the candid spec). While this mechanism allows to evolve services and clients
+independently without breaking them, it also means that a client may not receive all the information that the server is sending, e.g. in case the client schema lacks some fields that the server schema has.
+
+This loss of information is not an option for `ICRC-3`. The client must receive the same exact data the server sent. For this reason, `ICRC-3` introduces the `Value` type which never changes:
```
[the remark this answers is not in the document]
dskloetd: Where?

standards/ICRC-3/ICRC-3.did
```diff
@@ -0,0 +1,36 @@
+type Value = variant {
+    Blob : blob;
+    Text : text;
+    Nat : nat; // do we need this or can we just use Int?
+    Int : int;
+    Array : vec Value;
+    Map : vec record { text; Value };
+};
+
+type GetBlocksArgs = vec record { start : nat; length : nat };
```
sesi200, line 10:
```suggestion
type GetBlocksFn = record {
  canister_id : principal;
  method : text;
};

type GetBlocksArgs = vec record { start : nat; length : nat };
```

standards/ICRC-3/README.md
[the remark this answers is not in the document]
MarioDfinity: > The [candid](https://github.com/dfinity/candid) format supports sharing information even when the client and the server involved do not have the same schema (see the [Upgrading and subtyping](https://github.com/dfinity/candid/blob/master/spec/Candid.md#upgrading-and-subtyping) section of the candid spec). While this mechanism allows to evolve services and clients
> independently without breaking them, it also means that a client may not receive all the information that the server is sending, e.g. in case the client schema lacks some fields that the server schema has.
>
> This loss of information is not an option for `ICRC-3`. The client must receive the same exact data the server sent.

standards/ICRC-3/README.md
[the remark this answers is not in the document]
MarioDfinity: Fixed with 0e4b103.

standards/ICRC-3/README.md
MarioDfinity: Fixed with 25f86b7

standards/ICRC-3/README.md
MarioDfinity: Fixed with 1199b1f.

standards/ICRC-3/README.md
MarioDfinity: I changed this part, have a look again.

standards/ICRC-3/README.md
MarioDfinity: I think it makes sense to have it as part of the standard.

standards/ICRC-3/README.md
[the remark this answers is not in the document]
dskloetd: Yeah, that's the text that I commented on. It simply states that it's not acceptable, but it does not explain why.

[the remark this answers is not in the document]
MarioDfinity: I've added the following:

> This loss of information is not an option for `ICRC-3`. The client must receive the same exact data the server sent in order to verify it. Verification is done by hashing the data and checking that the result is consistent with what has been certified by the server.

Is it better?

standards/ICRC-3/README.md
MarioDfinity: I got rid of the schema language. People didn't like it :D.

standards/ICRC-3/README.md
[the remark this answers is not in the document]
dskloetd: Thanks. Why isn't it up to the client if it wants to verify the hashes? If it wants to verify the hashes, it should make sure to have an up-to-date schema. If it only wants certain data, it can use any schema that provides that data. By verifying the hash it can even make sure that its schema is up-to-date so.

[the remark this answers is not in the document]
MarioDfinity: The client cannot check that the schema is identical to the one of the canister. There is no guarantee that the schema will be identical even if the client first fetches it from the canister. Candid was built with a different goal in mind: the schema of the server can evolve independently from the client.
The solution to this problem is to have a type that will never change over time so it is guaranteed that clients and canisters will always know the full structure. That's what `Value` is.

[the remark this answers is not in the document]
roman-kashitsyn: > If it wants to verify the hashes, it should make sure to have an up-to-date schema.

Because it's not a sustainable solution in a decentralized system. If we went this way, releasing a new Ledger feature would immediately break block validation for all existing clients. We don't have a full list of our clients, so we cannot ask them to upgrade in advance.
The main clients of ICRC-3 standard are the on-chain and off-chain indexers, such as the index canister and Rosetta. We want these to continue working correctly with no intervention as we evolve the ledger.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,206 @@
...
+In other words the schema must be:
+
+```
+type ICRC1_Transaction = {
+    "phash" : Blob?;
+    "tx" : {
+        "op" : Text;
+        ...
+    };
+};
```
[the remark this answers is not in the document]
roman-kashitsyn: Because it's not a candid type, it's a schema of a Value object.

The `Value` type is like JSON, the schema here is like the JSON schema definition. It describes the shape of the value, which could not be expressed statically in Candid, so we had to invent a separate meta-language.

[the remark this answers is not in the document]
dskloetd: > Because it's not a candid type, it's a schema of a Value object.

That's exactly what I'm challenging.
OK, you need the `Value` type because you need to be able to express future inclusions.
But that doesn't mean that everything immediately needs to use the `Value` type.
Things that you already know must exist can be included in a proper type and the `Value` type can be used for future metadata.
Using this untyped type just makes things hard, so I'm suggesting to only use it where you have.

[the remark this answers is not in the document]
roman-kashitsyn: > untyped type

IMO, mixing "typed types" and "untyped types" will only create more confusion because validators will have to convert the data into a uniform "untyped" representation anyway to compute the block hash.
Providing a library for decoding typed records from Values is a simpler and cleaner design choice.

MarioDfinity:
@hpeebles the problem with that approach is that it makes the API more complex. Right now the input is a vector of `record { start : nat; length : nat }` which can potentially be used to go in any direction and work with any starting points. If the starting index is a optional last block seen and the direction is one more field then the input becomes a vector of `record { start : opt nat; length : nat; direction : Direction }`. This is way more powerful but adds complexity to the Ledger which I would like to avoid.

Instead we could have it as part of the Index standard. Even in your example, a client will want the transactions of a particular user which the Ledger cannot provide while an Index can. Indexes don't commit transactions nor deal with deduplication and can focus on more complex indexing and more powerful query API than Ledger.

standards/ICRC-3/README.md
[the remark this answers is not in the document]
MarioDfinity: I don't think the "untyped type" is harder to handle than the mixed one in this case. The problem with the mixed type is how to compute the block hash. I see two ways:
1. convert the mixed type to a "untyped type" but then you are basically writing the same code you need to go from "untyped type" to "typed type" just the other way around
2. Provide an algorithm that works on both "untyped type" and "typed type" and their mix (e.g. given something that resemble a structure, create a list of all fields hash, sort them and return the hash.). This would be as hard as the other two options if not harder because it's weirder. In your example, the metadata field should be "transparent" and its content should be treated as part of the "typed type"

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+ICRC-1 and ICRC-2 use the `tx` field to store input from the user and use the external block to store data set by the Ledger. For instance, the amount of a transaction is stored in the field `tx.amt` because it has been specified by the user, while the time when the block was added to the Ledger is stored in the field `ts` because it is set by the Ledger.
+
+A generic ICRC-1 or ICRC-2 Block:
+
+1. it MUST contain a field `ts: Nat` which is the timestamp of when the block was added to the Ledger
+2. if the `tx` field doesn't specify the fee then it MUST contain a field `fee: Nat` which specifies the fee payed to add this block to the Ledger
+3. its field `tx`
+    1. MUST contain a field `amt: Nat` that represents the amount
+    2. MUST contain the `fee: Nat` if the top-level `fee` is not set which is when the user didn't specify the expected `fee`
```
levifeldman, line 105:
```suggestion
    2. MUST contain the `fee: Nat` field if the user specifies the fee in the request. If the user does not specify the fee in the request, then this field is not set and the top-level `fee` is set.
```

levifeldman: More clear that if the user does specify the fee then this field is set.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,268 @@
...
+    3. CAN contain the `memo: Blob` field if specified by the user
+    4. CAN contain the `ts: Nat` field if specified by the user
```
levifeldman, line 107:
```suggestion
    4. CAN contain the `ts: Nat` field if the user sets the `created_at_time` field in the request.
```

MarioDfinity:
@levifeldman @bogwar @skilesare I've added `icrc3_get_archives`. Note that the API has pagination even if it's unlikely that the number of archives is going to be "too much" but better safe than sorry.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,288 @@
...
+2. specify the expected value of `tx.op`. This MUST be unique accross all the standards. `approve`, `burn`, `mint` and `xfer` are reserved for ICRC-1 and ICRC-2
+
+## [ICRC-1](../ICRC-1/README.md) and [ICRC-2](../ICRC-2/README.md) Block Schema
+
+ICRC-1 and ICRC-2 use the `tx` field to store input from the user and use the external block to store data set by the Ledger. For instance, the amount of a transaction is stored in the field `tx.amt` because it has been specified by the user, while the time when the block was added to the Ledger is stored in the field `ts` because it is set by the Ledger.
+
+A generic ICRC-1 or ICRC-2 Block:
+
+1. it MUST contain a field `ts: Nat` which is the timestamp of when the block was added to the Ledger
+2. if the `tx` field doesn't specify the fee then it MUST contain a field `fee: Nat` which specifies the fee payed to add this block to the Ledger
```
levifeldman, line 102:
```suggestion
2. if the operation requires a fee and if the `tx` field doesn't specify the fee then it MUST contain a field `fee: Nat` which specifies the fee payed to add this block to the Ledger
```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,288 @@
...
+3. its field `tx`
+    1. MUST contain a field `amt: Nat` that represents the amount
+    2. MUST contain the `fee: Nat` field if the user specifies the fee in the request. If the user does not specify the fee in the request, then this field is not set and the top-level `fee` is set.
```
levifeldman, line 105:
```suggestion
    2. MUST contain the `fee: Nat` field for operations that require a fee if the user specifies the fee in the request. If the user does not specify the fee in the request, then this field is not set and the top-level `fee` is set.
```

MarioDfinity: All ICRC-1 and ICRC-2 require a fee.

levifeldman: even mint and burn?

levifeldman: There is no fee field in the burn or mint examples.

levifeldman: Best to specify which operations require a fee field in icrc1/2 if mint and burn does not.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,288 @@
...
+A generic ICRC-1 or ICRC-2 Block:
+
+1. it MUST contain a field `ts: Nat` which is the timestamp of when the block was added to the Ledger
+2. if the operation requires a fee and if the `tx` field doesn't specify the fee then it MUST contain a field `fee: Nat` which specifies the fee payed to add this block to the Ledger
+3. its field `tx`
+    1. MUST contain a field `amt: Nat` that represents the amount
+    2. MUST contain the `fee: Nat` field if the user specifies the fee in the request. If the user does not specify the fee in the request, then this field is not set and the top-level `fee` is set.
+    3. CAN contain the `memo: Blob` field if specified by the user
+    4. CAN contain the `ts: Nat` field if the user sets the `created_at_time` field in the request.
+
```
levifeldman, line 108:
```suggestion

Operations that require paying a fee: Transfer, and Approve. 

```

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,290 @@
...
+2. it MUST contain a field `tx: Map`
+    1. `tx` MUST contain a field `op: Text`, aka operation, which uniquely describes the type of the Block. `op` values `approve`, `burn`, `mint` and `xfer` are reserved for ICRC-1 and ICRC-2 Blocks
+2. it MUST contain a field `phash: Blob` which is the [hash](#value-hash) of its parent if it has a parent block
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` MUST define the list of block schemas that it introduces. Each block schema MUST:
+
+1. extend the [Generic Block Schema](#generic-block-schema)
+2. specify the expected value of `tx.op`. This MUST be unique accross all the standards. `approve`, `burn`, `mint` and `xfer` are reserved for ICRC-1 and ICRC-2
```
dietersommer, line 93: The required uniqueness calls for a best practice that defines how a new ICRC standard should choose its op identifier in order to be unique; something like `icrc30xfer` or just `30xfer` could make sense. Obviously, ICRC-1 and ICRC 2 operations are special cases as those are implemented in production ledgers already.

Proposal:
"An `ICRC-x` standard MUST use namespacing for its op identifiers using the scheme `x || op-name` where `x` is the ICRC number and `op-name` is the name of the operation starting with a lowercase letter and comprising alphanumeric characters after the first character. For example, ICRC-30 would use `30approve`."

MarioDfinity: I pushed a change, can you have a look?

standards/ICRC-3/ICRC-3.did
```diff
@@ -0,0 +1,56 @@
...
+    canister_id : principal;
+
+    // The first block in the archive
+    start : nat;
+
+    // The last block in the archive
+    end : nat;
+}
+
+type GetBlocksArgs = vec record { start : nat; length : nat };
```
levifeldman, line 29: @MarioDfinity Do you think it is a good idea to paginate the archives over here, like in the icrc3_get_archives function? I can picture a common scenario where a client gets the latest block-id with get_tip_certificate and then queries for get_blocks with arg start: 0, len: tip_certificate.last_block_index which returns all the current blocks in the main-ledger and the archived_blocks with every archive canister and their ranges. Since you put this pagination on the icrc3_get_archives method, I'm thinking it might be a good idea to paginate the archives here too. What do you think?   

Feel free to change this suggestion however you feel is best. 

```suggestion
type GetBlocksArgs = vec record { start : nat; length : nat; from_archive: opt principal };
```

standards/ICRC-3/ICRC-3.did
```diff
@@ -0,0 +1,56 @@
...
+type GetArchivesResult = vec {
+    // The id of the archive
+    canister_id : principal;
+
+    // The first block in the archive
+    start : nat;
+
+    // The last block in the archive
+    end : nat;
+}
```
letmejustputthishere, line 27: i dont't think this is a valid candid type. is this supposed to be `vec record` or `record` instead?

letmejustputthishere, line 27:
```suggestion
type GetArchivesResult = vec {
    // The id of the archive
    canister_id : principal;

    // The first block in the archive
    start : nat;

    // The last block in the archive
    end : nat;
};
```
this is needed to for the candid to be valid i believe

standards/ICRC-3/ICRC-3.did
```diff
@@ -0,0 +1,56 @@
...
+    // block log
+    log_length : nat;
+
+    blocks : vec record { id : nat; block: Value };
+
+    archived_blocks : vec record {
+        args : GetBlocksArgs;
+        callback : func (GetBlocksArgs) -> (GetBlocksResult) query;
+    };
+};
```
letmejustputthishere, line 42: what happens if the response exceeds the message size limit? should there be something like the [`icrc7_max_take_value`](https://github.com/dfinity/ICRC/blob/icrc7-wg-draft/ICRCs/ICRC-7/ICRC-7.md#icrc7_max_take_value) for icrc3 as well?

standards/ICRC-3/ICRC-3.did
letmejustputthishere: applies to readme as well

standards/ICRC-3/ICRC-3.did
levifeldman: The get_blocks response already has a chunking mechanism for both, the blocks that the ledger itself contains and for the archives. If the archives don't fit into one response, then the ledger can return a callback to itself with the next range and then return the archives for that next range. So the above comment is not an issue.

levifeldman, line 29:
```suggestion
type GetBlocksArgs = vec record { start : nat; length : nat; account: opt Account };
```
This way ICRC-3 can be compatible with the index canister functionality.

standards/ICRC-3/ICRC-3.did
levifeldman: It might be better if this is specified in the spec.

dgdg-app:
Instead of "Value", something more-literal would help casual browsers of the Candid understand what is going on.

Something like "UpgradableType" or "Metatype"?

standards/ICRC-3/ICRC-3.did
MarioDfinity: Fixed with 8977901.

MarioDfinity: Fixed with 3c29c17

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,397 @@
...
+
+Validation of block `i` is done by checking the block hash against
+1. if `i + 1 < len(chain)` then the parent hash `phash` of the block `i+1`
+2. otherwise the `last_block_hash` in the tip certificate.
+
+## Generic Block Schema
+
+An ICRC-3 compliant Block
+
+1. MUST be a Value of variant Map
```
dietersommer, line 87: Please format the field names accordingly, e.g., `Value` of  variant `Map`, otherwise it is really not easy to read the specification.
Typo: fallsback -> falls back

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,397 @@
...
+2. MUST contain a field phash: Blob which is the hash of its parent if it has a parent block
+3. SHOULD contain a field type: String which uniquely describes the type of the Block. If this field is not set then the block type fallsback to ICRC-1 and ICRC-2 for backward compatibility purposes
+
+## Interaction with other standards
+
+Each standard that adheres to `ICRC-3` MUST define the list of block schemas that it introduces. Each block schema MUST:
+
+1. extend the [Generic Block Schema](#generic-block-schema)
+2. specify the expected value of `type`. This MUST be unique accross all the standards. An ICRC-x standard MUST use namespacing for its op identifiers using the scheme:
```
dietersommer, line 96: Maybe explain the scheme in words also, for example: "using the following scheme of using the ICRC standard's number as prefix to the name followed by an operation name that must begin with a letter:"

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,397 @@
...
+```
+op = icrc_number op_name
```
dietersommer, line 98: Let's use RFC 5234 syntax for formal languages consistently.
That would mean to use the forward slash `/` for separating options. We have currently a backslash `\` in `op_name`. Also, this is missing for the `nonzero_digit` and `digit` rules.
The `nonzero digit` in the second line misses an `_`.

standards/ICRC-3/README.md
```diff
@@ -0,0 +1,397 @@
...
+2. if the operation requires a fee and if the `tx` field doesn't specify the fee then it MUST contain a field `fee: Nat` which specifies the fee payed to add this block to the Ledger
+3. its field `tx`
+    1. MUST contain a field `amt: Nat` that represents the amount
+    2. MUST contain the `fee: Nat` field for operations that require a fee if the user specifies the fee in the request. If the user does not specify the fee in the request, then this field is not set and the top-level `fee` is set.
+    3. CAN contain the `memo: Blob` field if specified by the user
+    4. CAN contain the `ts: Nat` field if the user sets the `created_at_time` field in the request.
+
+Operations that require paying a fee: Transfer, and Approve.
+
+The type of a generic ICRC-1 or ICRC-2 Block is define by either the field `type` or the field `tx.op`. The first approach is preferred, the second one exists for backward compatibility.
```
dietersommer, line 127: "defined"

dietersommer (commented):
Made some comments on necessary fixes we should do.

standards/ICRC-3/README.md
MarioDfinity: I switched to RFC 523, can you double check what I wrote?

PR #130, opened 2023-07-14
feat(Fi-818) add SM and generic burn fn

NikolasHai:
This MR proposes the following changes:

1. Import the StateMachine and implement the test suite for it
2. Create generic burn functions to account for ledgers such as the cycles ledger
3. pump imports to align candid versions
4. tests for StateMachine


test/env/lib.rs
```diff
@@ -326,3 +473,76 @@ pub mod icrc1 {
...
+                .with_arg(Encode!(&()).unwrap())
+                .call()
+                .await?
+                .as_slice(),
+            Option<Account>
+        )
+        .unwrap()
+        .unwrap();
+        let res =
+            Decode!(agent.update(&canister_id, "icrc1_transfer").with_arg(Encode!(&Transfer::amount_to(amount,minting_account.owner)
```
MarioDfinity, line 496: Why you are passing just the owner?  You need to pass the full `minting_account` including the subaccount otherwise Ledgers that use Accounts with no default subaccount will get the wrong burn function, no?

test/env/lib.rs
```diff
@@ -326,3 +473,76 @@ pub mod icrc1 {
...
+).unwrap())
+.call_and_wait(waiter())
+.await.unwrap().as_slice(),Result<Nat, TransferError>)
+            .unwrap();
+        Ok(res)
+    })
+}
+
+pub fn standard_sm_burn_fn(
```
MarioDfinity, line 505: I fell like I'm missing something. If `LedgerEnv` provides `query`and `upgrade` then why there are two different functions for burn for replica and sm? You should be able to implement the standard burn function on top of `LedgerEnv`. What am I missing?

test/suite/lib.rs
```diff
@@ -63,74 +63,100 @@ async fn assert_balance(
...
+}
+
 /// Checks whether the ledger supports token transfers and handles
 /// default sub accounts correctly.
-pub async fn test_transfer(ledger_env: impl LedgerEnv + LedgerEnv) -> TestResult {
-    let receiver_env = ledger_env.fork();
-    let receiver = receiver_env.principal();
+/// Expects the given account to have a balance of at least 2*Transfer_Fee
+pub async fn test_transfer(ledger_env: impl LedgerEnv + LedgerTransaction + Clone) -> TestResult {
+    let envs = setup_test_accounts(&ledger_env, Nat::from(20_000), 2).await?;
```
MarioDfinity, line 102: Nitpick: you don't need to clone. Convert `envs` into an iterator, make it mutable and then call `next` twice.

test/suite/lib.rs
```diff
@@ -63,74 +63,100 @@ async fn assert_balance(
     Ok(())
 }
 
+async fn setup_test_accounts(
+    ledger_env: &(impl LedgerEnv + LedgerTransaction + std::clone::Clone),
+    amount: Nat,
+    acc_num: u64,
+) -> anyhow::Result<Vec<impl LedgerEnv + LedgerTransaction + Clone>> {
+    let mut res = vec![];
+    assert!(acc_num > 0);
```
MarioDfinity, line 72: Nitpick: the function still works also for `acc_num == 0`.

test/suite/lib.rs
```diff
@@ -63,74 +63,100 @@ async fn assert_balance(
     Ok(())
 }
 
+async fn setup_test_accounts(
```
MarioDfinity, line 66: This isn't nice to use because whatever you do you'll have to "unpack" the vector that is the result. Instead I would have a `setup_test_account` (singular). The usage would change from

```
let envs = setup_test_accounts(&ledger_env, amount, 2);
let env1 = envs[0].clone();
let env2 = envs[1].clone();
```

to

```
let env1 = setup_test_account(&ledger_env, amount);
let env2 = setup_test_account(&ledger_env, amount);
```

The additional bonus is that it's more flexible allowing to create test accounts with different amount of tokens.

test/suite/lib.rs
```diff
@@ -63,74 +63,100 @@ async fn assert_balance(
...
+    let burn_amount = Nat::from(10_000);
+    let envs = setup_test_accounts(&ledger_env, burn_amount.clone(), 1).await?;
+    let p1_env = envs[0].clone();
+
+    p1_env
+        .burn(burn_amount.clone())
+        .await?
+        .context("failed to burn amount")?;
+
+    assert_balance(&p1_env, p1_env.principal(), 0).await?;
```
MarioDfinity, line 159: Shouldn't this be `previous balance - burn_amount` instead of `0`? You don't know if the account had tokens in it. It's unlikely but not impossible. We can use a generic invariant that works regardless of the balance of the account before the `setup_test_accounts`.

test/env/lib.rs
```diff
@@ -162,14 +164,20 @@ pub trait LedgerEnv {
         Output: for<'a> ArgumentDecoder<'a>;
 }
 
+pub type BurnReturnType =
+    Pin<Box<dyn std::future::Future<Output = anyhow::Result<Result<Nat, TransferError>>>>>;
+
+pub type ReplicaBurnFn = fn(Arc<Agent>, Principal, Nat) -> BurnReturnType;
```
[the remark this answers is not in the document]
NikolasHai: The ReplicaBurnFn is different to the StateMachineBurnFn, so the name LedgerBurnFn would be ambiguous, no?

test/ref/test.rs
```diff
@@ -99,8 +118,56 @@ async fn install_canister(agent: &Agent, wasm: &[u8], init_arg: &[u8]) -> Princi
     canister_id
 }
 
-#[tokio::main]
-async fn main() {
+fn unzip_file(file_path: &str, output_path: &str) {
```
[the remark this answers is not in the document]
NikolasHai: I tried to do it with bazel first but I didn't manage. The two ways i tried were:
- Use https_archive: Unzips the file but it cannot be set to executable. To make it executable I tried gen_rule. I could however not use gen_rule since the bazel directory is read-only. So the only way would have been to create a new directory in bazel, copy the unzipped binary to that directory and then make it executable through a file group. 
- Use https_file: The file can be set to executable but is not unzipped. Unzipping the file poses the same issue as with https_archive which is that the bazel directory is read-only. 

I opted for this solution as it seemed the most strait forward. Maybe there is an easier solution with bazel that I did not find?

test/env/lib.rs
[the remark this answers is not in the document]
MarioDfinity: I'm missing another part then. I was expecting the state machine structure to be the same as the replica one with the Agent changed from the one that talks to a replica to the one that talks to the StateMachine process. I need reread again how we use SM.

test/env/lib.rs
NikolasHai: It depends on how the burn function is set up. For example for the cycles ledger the process of burning tokens requires more than just query and update calls. Originally we also intended to use the statemachine to be able to mint tokens/cycles too, which would definitely require more than just query and update calls.

test/ref/test.rs
```diff
@@ -99,8 +118,56 @@ async fn install_canister(agent: &Agent, wasm: &[u8], init_arg: &[u8]) -> Princi
...
+    );
+
+    // Get the current permissions
+    let mut permissions = fs::metadata(new_path.clone()).unwrap().permissions();
+
+    // Add the executable permission
+    permissions.set_mode(permissions.mode() | 0o111);
+
+    // Set the updated permissions
+    fs::set_permissions(new_path.clone(), permissions).unwrap();
```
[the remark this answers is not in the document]
NikolasHai: Same explanation as before.

test/suite/lib.rs
NikolasHai: When setting up the testing account, I am checking whether the newly generated account has a balance of 0, so before this assert, I can be sure that the account had 0 tokens at the beginning of the test. I then transfer 10_000 tokens to it and burn them again. That is why I check for a balance 0. Of course, it is theoretically possible that the random number generator generates an account which already exists and has non zero balance or that it is an active account and transfers are being made while I a running the test, but I would categorize this as a very low possibility as the numeric space of random numbers for u64 is pretty large.

test/env/lib.rs
MarioDfinity: I see but wouldn't the `standard_*_burn_fn` be the same? Then the cycles ledger won't use the standard function.

test/ref/test.rs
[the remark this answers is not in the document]
MarioDfinity: Can't we do the same thing we do for replica, sandbox and starter https://sourcegraph.com/github.com/dfinity/ICRC-1/-/blob/bazel/replica_tools.bzl ?

test/suite/lib.rs
```diff
@@ -63,74 +63,100 @@ async fn assert_balance(
...
+    Ok(res)
+}
+
 /// Checks whether the ledger supports token transfers and handles
 /// default sub accounts correctly.
-pub async fn test_transfer(ledger_env: impl LedgerEnv + LedgerEnv) -> TestResult {
-    let receiver_env = ledger_env.fork();
-    let receiver = receiver_env.principal();
+/// Expects the given account to have a balance of at least 2*Transfer_Fee
+pub async fn test_transfer(ledger_env: impl LedgerEnv + LedgerTransaction + Clone) -> TestResult {
```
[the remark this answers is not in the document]
MarioDfinity: `trait LedgerEnv: Clone { ... }` ?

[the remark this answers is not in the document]
NikolasHai: Of course...

test/env/lib.rs
NikolasHai: maybe I can make this work. I'll get back to this.

NikolasHai: Just using the trait for the `standard burn function` is not "object safe" when used as a field in the implementations of `LedgerEnv` so I had to find a work around. The implementations of `LedgerEnv` are now using a the same standard burn function which only relies on the trait `LedgerEnv`

test/suite/lib.rs
```diff
@@ -63,74 +63,91 @@ async fn assert_balance(
     Ok(())
 }
 
-/// Checks whether the ledger supports token transfers and handles
-/// default sub accounts correctly.
-pub async fn test_transfer(ledger_env: impl LedgerEnv + LedgerEnv) -> TestResult {
+async fn transfer_or_fail(
```
MarioDfinity, line 66: Annotate with `#[track_caller]`

test/suite/lib.rs
```diff
@@ -63,74 +63,91 @@ async fn assert_balance(
...
+
+async fn setup_test_account(
+    ledger_env: &(impl LedgerEnv + LedgerTransaction + std::clone::Clone),
+    amount: Nat,
+) -> anyhow::Result<impl LedgerEnv + LedgerTransaction> {
+    let balance = balance_of(ledger_env, ledger_env.principal()).await?;
+    assert!(balance >= amount.clone() + transfer_fee(ledger_env).await?);
     let receiver_env = ledger_env.fork();
     let receiver = receiver_env.principal();
+    assert_balance(&receiver_env, receiver, 0).await?;
```
MarioDfinity, line 85: While you are at it can you also annotate `assert_balance` with `#[track_caller]`?

test/ref/test.rs
```diff
@@ -99,8 +118,56 @@ async fn install_canister(agent: &Agent, wasm: &[u8], init_arg: &[u8]) -> Princi
...
+
+        I looked for it at {:?}. You can specify another path with the environment variable STATE_MACHINE_BINARY (note that I run from {:?}).
+
+        Run the following command to get the binary:
+            curl -sLO https://download.dfinity.systems/ic/$commit/binaries/$platform/ic-test-state-machine.gz
+            gzip -d ic-test-state-machine.gz
+            chmod +x ic-test-state-machine
+        where $commit can be read from `.ic-commit` and $platform is 'x86_64-linux' for Linux and 'x86_64-darwin' for Intel/rosetta-enabled Darwin.
+        ", &path, &env::current_dir().map(|x| x.display().to_string()).unwrap_or_else(|_| "an unknown directory".to_string()));
+    }
```
roman-kashitsyn: These instructions are incorrect for our case; we need to supply the binary via Bazel.

PR #140, opened 2023-07-27
feat(FI-858) test for deduplication

NikolasHai:
This MR adds a deduplication test for the ICRC-1 test suite


test/suite/lib.rs
```diff
@@ -218,13 +219,80 @@ pub async fn test_supported_standards(ledger: impl LedgerEnv) -> anyhow::Result<
...
+            < transfer(&p1_env, transfer_args.clone())
+                .await
+                .unwrap()
+                .unwrap()
+    );
+
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
```
roman-kashitsyn, line 244: That code might not work with the StateMachine tests.
We need `current_time()` or something like that in LedgerEnv.

test/suite/lib.rs
```diff
@@ -218,13 +219,80 @@ pub async fn test_supported_standards(ledger: impl LedgerEnv) -> anyhow::Result<
...
+    assert!(
+        transfer(&p1_env, transfer_args.clone())
+            .await
+            .unwrap()
+            .unwrap()
+            < transfer(&p1_env, transfer_args.clone())
+                .await
+                .unwrap()
+                .unwrap()
+    );
```
roman-kashitsyn, line 239: We must not panic in the test suite because this will crash the entire test process. We should return an `Outcome` instead.

test/suite/lib.rs
NikolasHai: You were right, I fixed it

test/suite/lib.rs
```diff
@@ -218,13 +219,80 @@ pub async fn test_supported_standards(ledger: impl LedgerEnv) -> anyhow::Result<
     Ok(Outcome::Passed)
 }
 
+/// Checks whether the ledger advertizes support for ICRC-1 standard.
```
maciejdfinity, line 222: comment copy paste :)

test/suite/lib.rs
```diff
@@ -218,13 +219,84 @@ pub async fn test_supported_standards(ledger: impl LedgerEnv) -> anyhow::Result<
...
+        .await
+        .unwrap()
+        .unwrap();
+
+    // If we send the transfer again it should be a duplicate --> Transfer should not succeed
+    assert!(transfer(&p1_env, transfer_args.clone())
+        .await
+        .unwrap()
+        .is_err());
+
```
maciejdfinity, line 285: We could also test whether the ledger correctly identifies that transfer to (principal, None) and (principal, Some([0; 32])) are different transactions and are not deduplicated.

NikolasHai: Good idea

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-@dfinity/finint
+@dfinity/finint
```
roman-kashitsyn, line 1: Why did we change this file?

roman-kashitsyn (approved):
LGTM, thanks!

PR #191, opened 2024-07-24
list allowances standard initial commit

bogwar:
This is a draft for the ICRC-191 standard for listing outstanding allowances on ICP ledgers

### DIFF
--- a/standards/ICRC-191/README.md
+++ b/standards/ICRC-191/README.md
@@ -52,19 +52,22 @@ type Allowance = record {
 }
 ```
 
+The endpoint returns up to `taken` allowances of the from_account.owner, starting with the allowance between `from_account` and `to_account`.
+
 
 ## 4. Semantics
 
-Recall that outstanding allowances are (per the ICRC-2 standard) specifdied as map from pairs of accounts, to allowances.  To specify the behavior of icrc191_list_allowances we make the assumption that the set of pairs (Account, Account) is ordered, lexicographically.
-Let first_principal be the first principal, in lexicographically, and first_subaccount be the first subaccount, lexicographically.
-
-If from_account is not specified, then this is instantiated with Account{caller_principal, first_subaccount}.  
-If prev_spender is not specified, then this is instantiated with Account{first_principal, first_subaccount}.
-
-The endpoint returns the list of records of the form (account_1, account_2, allowance), in lexicographic order starting with the allowance of (from_account,prev_spender). The list contains at most `taken` entries.
+Recall that outstanding allowances are (per the ICRC-2 standard) specified as a map from pairs of accounts, to allowances.  To specify the behavior of icrc191_list_allowances we require that the set of pairs (Account, Account) is ordered, lexicographically.
+Let `first_principal` be the lexicographically first principal `first_subaccount` be the lexicographically first subaccount (this is the default subaccount, i.e. the all-0 32 byte string).
+Let `caller_principal` be the principal of the caller.
 
 
+If `from_account` is not provided, then this is instantiated with `Account{caller_principal, first_subaccount}`.  
+If `from_account.subaccount` is not specified then this is instantiated with `first_subaccount`.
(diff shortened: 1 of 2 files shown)


standards/ICRC-191/README.md
```diff
@@ -0,0 +1,90 @@
...
+    subaccount : opt blob;
+};
+
+type Allowance = record {
+  allowance : nat;
+  expires_at : opt nat64;
+}
+```
+
+The endpoint returns up to `taken` allowances of the from_account.owner, starting with the allowance between `from_account` and `to_account`.
```
skilesare, line 55: take

standards/ICRC-191/README.md
```diff
@@ -0,0 +1,90 @@
+| Status |
+|:------:|
+|Draft|
+
+# ICRC-191: Enhanced Allowance Query Mechanism with Pagination
```
skilesare, line 5: typically the number comes from github.com/dfinity/icrc and not from the icrc-1 repo...I think we're around about 101 or 102.

standards/ICRC-191/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+The endpoint returns up to `taken` allowances of the from_account.owner, starting with the allowance between `from_account` and `to_account`.
+
+
+## 4. Semantics
+
+Outstanding allowances, as specified in the ICRC-2 standard, are represented as a map from pairs of accounts to allowances. To specify the behavior of `icrc191_list_allowances`, the set of pairs `(Account, Account)` is ordered lexicographically. Let `first_principal` be the lexicographically first principal, and `first_subaccount` be the lexicographically first subaccount (the default subaccount, i.e., the all-0 32-byte string). Let `caller_principal` be the principal of the caller.
+
+The `icrc191_list_allwances` method behaves as follows:
```
skilesare, line 62: spelling

standards/ICRC-191/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+- A1 = ((p0,s0), (p1,s1), a1)
+- A2 = ((p0,s0), (p2,s2), a2)
+- A3 = ((p0,s1), (p3,s3), a3)
+- A4 = ((p1,s1), (p4,s4), a4)
+- A5 = ((p1,s2), (p5,s5), a5)
+
+Then:
+
+- If `p0` calls the list allowances endpoint, with `from_account = (p0, s0)`, `prev_spender = None` and `take=4` the endpoint returns ``(A1, A2, A3)``, i.e. the endpoint only returns allowances of accounts belonging to `p0`.  It is limited to allowances of `p0`, but not only to those having `(p0,s0)` as source of the allowance.
```
skilesare, line 86: I think you need an example of from_account = (p0, s1) (which I think would just get you the one item because this is in lexicographic order correct?

I still think I like 

```
type ListAllowancesArgs = record {
    from_account : opt { 
        owner = principal;
        subaccount = opt opt blob;
    };
    prev_spender : opt Account;
    take : opt nat;
}
```

....so that I can indicate I either want All subaccount for a principal with null or A specific subaccount with opt null for default subaccount and opt opt blob for a specific subaccount.

But I seem to recall you may have had a really good reason not to do this?  I apologize...the meeting was a while back and I don't quite remember the note.

skilesare (changes requested):
A few spelling or technical things(like the number)...one philosophical difference, but it likely fine either way.

standards/ICRC-191/README.md
bogwar: I think the API is already not very intuitive so I'd lean towards pushing the implied complexity related to result filtering on the users of the API rather than on the implementation of the API itself. WDYT?

standards/ICRC-191/README.md
bogwar: got it

standards/ICRC-191/README.md
bogwar: moved it there: got number 103

PR #196, opened 2024-11-28
ICRC-106: standard for discovering the location of the index canister from the ledger canister

bogwar:
A standard to allow for the discovery of the index canister from the corresponding ledger. 
A secondary goal for the standard is to document the interface of the current index canister.
This standard will exist in draft form only since we will work on a proper ICRC standard for index canisters (which will include discoverability).

### DIFF
--- a/standards/ICRC-106/ICRC-106.md
+++ b/standards/ICRC-106/ICRC-106.md
@@ -1,0 +1,241 @@
+| Status |
+|:------:|
+|Draft|
+
+# ICRC-106: Standard for Associating Index Canisters with ICRC-1 Tokens
+
+## 1. Introduction
+
+Wallet applications and token management tools often need to retrieve both token metadata and transaction history for a given principal. Index canisters provide such data, but identifying an associated index canister for ICRC-1 tokens is currently not standardized, leading to inconsistencies in wallet integrations.
+
+Standard **ICRC-106** :
+1. Introduces a standard approach for indicating the presence of an index canister for ICRC-1 tokens through ledger metadata.
+2. Defines a minimal interface for the associated index canister to facilitate querying transaction history in a consistent manner.
+
+This draft standard aims to improve interoperability, simplify wallet integrations, and enable token-related applications to reliably access transaction histories.  It acts as a placeholder and documentation source for the API of the index canister, until a comprehensive standard for index canisters will be developed.
+
+
+## 2. Metadata
+
+A ledger implementing ICRC-106 MUST include the following entry in the output of the `icrc1_supported_standards` method:
+
+```candid
+record { name = "ICRC-106"; url = "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-106" }
+```
+
+Additionally, the ledger MUST provide the following metadata entry retrievable via the `icrc1_metadata` method:
+
+- `icrc106:index_principal` (text): The textual representation of the principal of the associated index canister.
+
+These metadata entries allow clients to discover and interact with the index canister associated with a ledger and can be retrieved using method `icrc1_metadata` defined by the ICRC-1 standard.
+
+
+Compliant ledgers MUST also implement the following endpoint for programmatically retrieving the index principal:
+
+```candid
+
+icrc106_get_index_principal: () -> GetIndexPrincipalsResult query;
+
+type GetIndexPrincipalResult = variant {
+    Ok : principal;
+    Err : GetIndexPrincipalError;
+};
+
+type GetIndexPrincipalError = variant {
+    IndexPrincipalNotSet;
+
+    // Any error not covered by the above variants.
+    GenericError: record {
+       error_code: nat;
+       description: text;
+   };
+};
+
+```
+
+The metadata entry `icrc106:index_principal` and the `icrc106_get_index_principal` method MUST provide consistent information. Specifically:
+
+- The `icrc106:index_principal` metadata entry MUST represent the textual form of the principal returned by the `icrc106_get_index_principal` method.  
+- The `icrc106_get_index_principal` method MUST return the principal corresponding to the index canister associated with the ledger, as specified in the `icrc106:index_principal` metadata entry.
+
+This requirement ensures that both mechanisms reliably point to the same index canister.
+
+
+## 3. Index Canister Interface
+
+The index canister associated with the ledger SHOULD implement the following minimal Candid interface:
+
+```candid
+type Tokens = nat;
+
+type BlockIndex = nat;
+
+type SubAccount = blob;
+
+type Account = record {
+    owner: principal;
+    subaccount: opt SubAccount;
+};
+
+type GetAccountTransactionsArgs = record {
+    account: Account;
+    start: opt BlockIndex; // The block index of the last transaction seen by the client.
+    max_results: nat; // Maximum number of transactions to fetch.
+};
+
+type Burn = record {
+    from : Account;
+    memo : opt blob;
+    created_at_time : opt nat64;
+    amount : Tokens;
+    spender : opt Account;
+};
+
+type Mint = record {
+  to : Account;
+  memo : opt blob;
+  created_at_time : opt nat64;
+  amount : Tokens;
+};
+
+type Transfer = record {
+    to : Account;
+    fee : opt Tokens;
+    from : Account;
+    memo : opt blob;
+    created_at_time : opt nat64;
+    amount : Tokens;
+    spender : opt Account;
+};
+
+type Approve = record {
+    fee : opt Tokens;
+    from : Account;
+    memo : opt blob;
+    created_at_time : opt nat64;
+    amount : Tokens;
+    expected_allowance : opt nat;
+    expires_at : opt nat64;
+    spender : Account;
+};
+
+type Transaction = record {
+    burn : opt Burn;
+    kind : text;
+    mint : opt Mint;
+    approve : opt Approve;
+    timestamp : nat64;
+    transfer : opt Transfer;
+};
+
+type TransactionWithId = record {
+    id : BlockIndex;
+    transaction : Transaction;
+};
+
+type GetTransactions = record {
+  balance : Tokens;
+  transactions : vec TransactionWithId;
+  // The txid of the oldest transaction the account has
+  oldest_tx_id : opt BlockIndex;
+};
+
+type GetTransactionsErr = record {
+  message : text;
+};
+
+type GetAccountTransactionsResult = variant {
+  Ok : GetTransactions;
+  Err : GetTransactionsErr;
+};
+
+type ListSubaccountsArgs = record {
+    owner: principal;
+    start: opt SubAccount;
+};
+
+type Status = record {
+    num_blocks_synced : BlockIndex;
+};
+
+service : {
+    get_account_transactions: (GetAccountTransactionsArgs) -> (GetAccountTransactionsResult) query;
+    ledger_id: () -> (principal) query;
+    list_subaccounts : (ListSubaccountsArgs) -> (vec SubAccount) query;
+    status : () -> (Status) query;
+}
+```
+
+
+
+# Methods Provided by the Index Canister
+
+The index canister provides methods to facilitate querying of transaction history and metadata associated with accounts. Below is a description of the relevant methods specified in this standard, including their purpose, input, output, and typical use case.
+
+---
+
+## get_account_transactions
+- **Purpose**: Retrieves transactions associated with a specific account, starting from a specified block index and returning up to a maximum number of results. Transactions are returned in **reverse chronological order** (newest to oldest).
+- **Input**:
+  - `account`: The account (principal and optional subaccount) for which transactions are to be fetched.
+  - `start`: *(Optional)* The block index of the most recent transaction the client has already seen. If provided, only transactions with block indices **less than** this value will be returned.
+  - `max_results`: The maximum number of transactions to return.
+- **Output**:
+  - **`Ok`**: Returns a record containing:
+    - `balance`: The current token balance of the account.
+    - `transactions`: A vector of `TransactionWithId`, each containing:
+      - `id`: The block index of the transaction.
+      - `transaction`: Details of the transaction (burn, transfer, approval, and timestamp).
+    - `oldest_tx_id`: *(Optional)* The block index of the oldest transaction for the account, or `None` if no transactions exist.
+- **Typical Use Case**: This method is often used by wallets to display transaction history and update account balances. It also supports paginated transaction retrieval for efficient history browsing.
+
+---
+
+## list_subaccounts
+- **Purpose**: Lists all subaccounts associated with a specified principal.
+- **Input**:
+  - `owner`: The principal for which to list subaccounts.
+  - `start`: *(Optional)* Specifies the subaccount to start listing from. Only subaccounts lexicographically greater than `start` will be included. If start is omitted, the method will return all subaccounts from the beginning of the list, ordered lexicographically.
+- **Output**: A vector of `SubAccount`, each representing a subaccount under the specified principal. The list will be empty if the principal has not used subaccounts, or if there are no subaccounts lexicographically higher than `start`.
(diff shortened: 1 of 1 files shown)


standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+A ledger implementing ICRC-106 MUST include the following entry in the output of the `icrc1_supported_standards` method:
+
+```candid
+record { name = "ICRC-106"; url = "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-106" }
+```
+
+Additionally, the ledger MUST provide the following metadata entry retrievable via the `icrc1_metadata` method:
+
+- `icrc106:index_principal` (text): The textual representation of the principal of the associated index canister.
```
mraszyk, line 28: Is there no principal type available (so that one has to resort to the textual representation)?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+These metadata entries allow clients to discover and interact with the index canister associated with a ledger and can be retrieved using method `icrc1_metadata` defined by the ICRC-1 standard.
+
+
+Compliant ledgers MUST also implement the following endpoint for programmatically retrieving the index principal:
+
+```candid
+
+icrc106_get_index_principal: () -> (GetIndexPrincipalsResult) query;
```
mraszyk, line 37:
```suggestion
icrc106_get_index_principal: () -> (GetIndexPrincipalResult) query;
```

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+   };
+};
+
+```
+
+The metadata entry `icrc106:index_principal` and the `icrc106_get_index_principal` method MUST provide consistent information. Specifically:
+
+- The `icrc106:index_principal` metadata entry MUST represent the textual form of the principal returned by the `icrc106_get_index_principal` method.  
+- The `icrc106_get_index_principal` method MUST return the principal corresponding to the index canister associated with the ledger, as specified in the `icrc106:index_principal` metadata entry.
+- If the entry `icrc106:index_principal` does not exist or is not appropriately set, the endpoint should return `IndexPrincipalNotSet`.
```
mraszyk, line 60: what does "appropriately" mean in this context?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+       description: text;
+   };
+};
+
+```
+
+The metadata entry `icrc106:index_principal` and the `icrc106_get_index_principal` method MUST provide consistent information. Specifically:
+
+- The `icrc106:index_principal` metadata entry MUST represent the textual form of the principal returned by the `icrc106_get_index_principal` method.  
+- The `icrc106_get_index_principal` method MUST return the principal corresponding to the index canister associated with the ledger, as specified in the `icrc106:index_principal` metadata entry.
```
mraszyk, line 59: This condition seems to be redundant given the previous one.

standards/ICRC-106/ICRC-106.md
mraszyk: In any case, the first condition "The `icrc106:index_principal` metadata entry MUST represent the textual form of the principal returned by the `icrc106_get_index_principal` method. " seems to be sufficient to ensure consistency.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+  Err : GetTransactionsErr;
+};
+
+type ListSubaccountsArgs = record {
+    owner: principal;
+    start: opt SubAccount;
+};
+
+type Status = record {
+    num_blocks_synced : BlockIndex;
```
mraszyk, line 159: there seems to be a type mismatch: the prefix num_ suggests a number (count) while BlockIndex suggests an index (sequence) number of a block index.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+};
+
+type GetAccountTransactionsResult = variant {
+  Ok : GetTransactions;
+  Err : GetTransactionsErr;
+};
+
+type ListSubaccountsArgs = record {
+    owner: principal;
+    start: opt SubAccount;
```
mraszyk, line 155: The field start seems to aim at supporting pagination, but the return type of list_subaccounts is only a vector without pagination information.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+type Status = record {
+    num_blocks_synced : BlockIndex;
+};
+
+service : {
+    get_account_transactions: (GetAccountTransactionsArgs) -> (GetAccountTransactionsResult) query;
+    ledger_id: () -> (principal) query;
+    list_subaccounts : (ListSubaccountsArgs) -> (vec SubAccount) query;
+    status : () -> (Status) query;
```
mraszyk, line 166: all these endpoints are queries: are they gonna be certified?

standards/ICRC-106/ICRC-106.md
mraszyk: we're also lacking a field to bound the maximum number of subaccounts to get in the response

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+type TransactionWithId = record {
+    id : BlockIndex;
+    transaction : Transaction;
+};
+
+type GetTransactions = record {
+  balance : Tokens;
+  transactions : vec TransactionWithId;
+  // The txid of the oldest transaction the account has
+  oldest_tx_id : opt BlockIndex;
```
mraszyk, line 141: Why is this useful?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+    transfer : opt Transfer;
+};
+
+type TransactionWithId = record {
+    id : BlockIndex;
+    transaction : Transaction;
+};
+
+type GetTransactions = record {
+  balance : Tokens;
```
mraszyk, line 138: Wouldn't it make sense to expose the balance via a dedicated method: currently, it is not clear what the relationship is between this balance and the (subset of) transactions in the subsequent field.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+    from : Account;
+    memo : opt blob;
+    created_at_time : opt nat64;
+    amount : Tokens;
+    expected_allowance : opt nat;
+    expires_at : opt nat64;
+    spender : Account;
+};
+
+type Transaction = record {
```
mraszyk, line 123: Why are the field burn, mint, approve, transfer optional instead of having a single field of an enumeration type?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+    burn : opt Burn;
+    kind : text;
```
mraszyk, line 125: Is this the textual represenation of the variant (option) included? If so, a variant type would be more consistent.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+    created_at_time : opt nat64;
+    amount : Tokens;
+    spender : opt Account;
+};
+
+type Mint = record {
+  to : Account;
+  memo : opt blob;
+  created_at_time : opt nat64;
+  amount : Tokens;
```
mraszyk, line 99: these three fields seem to be common for every operation (Burn, Mint, Transfer, Approve) and thus could be factored out

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+    get_account_transactions: (GetAccountTransactionsArgs) -> (GetAccountTransactionsResult) query;
+    ledger_id: () -> (principal) query;
+    list_subaccounts : (ListSubaccountsArgs) -> (vec SubAccount) query;
+    status : () -> (Status) query;
+}
+```
+
+
+
+# Methods Provided by the Index Canister
```
mraszyk, line 172: Shouldn't this section be placed before the candid interface? (So that the description comes first and the reader is not confused by hitting an undocumented candid declarations first.)

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+---
+
+## status
+- **Purpose**: Retrieves the synchronization and operational status of the index canister.
+- **Input**: None.
+- **Output**: A `Status` record containing:
+  - `num_blocks_synced`: The total number of blocks that have been successfully synchronized by the index canister.
+- **Typical Use Case**: Used for monitoring the health and synchronization status of the index, this method is helpful for determining whether the index has fully caught up with the ledger and is operational.
```
mraszyk, line 220: What does it mean to be "fully caught up"? One would need to query the number of blocks in the ledger and index which are two separate async calls.

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+## Optional Methods
+
+While the methods defined in this standard are sufficient for compliance with ICRC-106, certain implementations of the index canister may include additional methods to extend functionality. These methods are not required by ICRC-106 but may be present for advanced use cases:
+
+- **`get_blocks`**: Fetches raw block data for a specified range of indices. This is useful for applications requiring detailed historical data.
```
mraszyk, line 227: > for a specified range of indices

isn't this paginated with the response potentially less than the specified amount of block indices?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+## status
+- **Purpose**: Retrieves the synchronization and operational status of the index canister.
+- **Input**: None.
+- **Output**: A `Status` record containing:
+  - `num_blocks_synced`: The total number of blocks that have been successfully synchronized by the index canister.
+- **Typical Use Case**: Used for monitoring the health and synchronization status of the index, this method is helpful for determining whether the index has fully caught up with the ledger and is operational.
+
+
+## Optional Methods
```
mraszyk, line 223: How useful is this section given that it doesn't really provide much details on the method listed?

standards/ICRC-106/ICRC-106.md
```diff
@@ -0,0 +1,242 @@
...
+
+This draft standard aims to improve interoperability, simplify wallet integrations, and enable token-related applications to reliably access transaction histories.  It acts as a placeholder and documentation source for the API of the index canister, until a comprehensive standard for index canisters will be developed.
+
+
+## 2. Metadata
+
+A ledger implementing ICRC-106 MUST include the following entry in the output of the `icrc1_supported_standards` method:
+
+```candid
+record { name = "ICRC-106"; url = "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-106" }
```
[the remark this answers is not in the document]
bogwar: This is the location of where the standard once the PR is merged.  In the meantime there was some reorganisation of the working gorups (and of repositories) so there will be a new location. I'll leave this open so that I remember to update the link accordingly.

standards/ICRC-106/ICRC-106.md
bogwar: Not really; metadata can only be some basic type (per ICRC-1 [metadata standard](https://internetcomputer.org/docs/current/references/icrc1-standard#icrc1_metadata-))

standards/ICRC-106/ICRC-106.md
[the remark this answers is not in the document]
bogwar: This was a request from the working group (i think it was aligned with some other standards (iirc ICRC-37).

standards/ICRC-106/ICRC-106.md
bogwar: Good point;  a ledger that does not obey the first requirement (in which case it's already not compliant so it doesn't really matter what the endpoint returns)

standards/ICRC-106/ICRC-106.md
bogwar: There is a bit more information in the second part (since it also talks about the index canister) but yes, it can be consolidated into a single blurb

standards/ICRC-106/ICRC-106.md
bogwar: For most of the remaining questions/suggestions the answer is: the standard documents the interface of the existing index canister and there are no plans to change that. 

The bigger context is that the plan is to keep this standard as draft to serve two purposes. One is to document the methods discussed above and two is to  serve as documentation for the existing index which cannot be standardised:  a good practice that we adhered to in previous standards is to prefix standardised  method names with the name of the standard. 
We will work on a new standard for indexes at some point in the near future.

standards/ICRC-106/ICRC-106.md
bogwar: see above

standards/ICRC-106/ICRC-106.md
bogwar: For most of the remaining questions/suggestions (including this one) the answer is: the standard documents the interface of the existing index canister and there are no plans to change that since it's already being rather heavily used. 

The bigger context is that the plan is to keep this standard as draft to serve two purposes. One is to document the methods discussed above and two is to  serve as documentation for the existing index which cannot be standardised:  a good practice that we adhered to in previous standards is to prefix standardised  method names with the name of the standard. 
We will work on a new standard for indexes at some point in the near future so I'll keep these points open since they're useful to take into account then.

standards/ICRC-106/ICRC-106.md
bogwar: The index will always be chasing after the ledger but the usecase here refers to the situation when there's something wrong and the ledger is significantly behind its target.  I think it was inspired by Rosetta which has a similar endpoint.

standards/ICRC-106/ICRC-106.md
bogwar: From the point of view of documenting the existing index it says that "there's these other bunch of methods implemented by the index"

standards/ICRC-106/ICRC-106.md
bogwar: I think it's not paginated but there's a strict bound on how many blocks you can get in a call. I'll double check

bogwar: there's a 2000 blocks upperbound so one doesn't get paginated responses on this endpoint.

standards/ICRC-106/ICRC-106.md
mraszyk: A blob would be alternative, but a textual representation is likely easier to understand.

standards/ICRC-106/ICRC-106.md
mraszyk: > there's a 2000 blocks upperbound so one doesn't get paginated responses on this endpoint.

2000 blocks in a single response could make this endpoint very expensive to process

standards/ICRC-106/ICRC-106.md
mraszyk: > see above

I'm not sure what you refer to. Could you please provide a link?

standards/ICRC-106/ICRC-106.md
mraszyk: Indeed, but if the input/output types and their description is omitted, then how is a reader suppose to use this information?

standards/ICRC-106/ICRC-106.md
bogwar: Right. What I ment is that the call doesn't get a paginated response but rather a set of blocks specified by the bounds in the call, restricted to some constant (currently 2000, we'll most likely shrink that as part of the expensive endpoints epic in FI).

standards/ICRC-106/ICRC-106.md
bogwar: They could look the at the .did file of the index canister.  But I think this discussion points out to simply omitting this part and not even popularising those methods anyway

standards/ICRC-106/ICRC-106.md
bogwar: Sorry about the fuzzy pointer. I was referring to the clarification that I've added about the standard documenting the index interface (it's somewhere below :) )

PR #199, opened 2025-08-06
ICRC-3 refinement

bogwar:
Proposed change to ICRC-3 based on [this proposal](https://docs.google.com/document/d/19jl7FX-TcNhI0EafPkBX7lyhesHq6VAMrtfdaChzcDQ/edit?usp=sharing)

### DIFF
--- a/standards/ICRC-3/README.md
+++ b/standards/ICRC-3/README.md
@@ -153,23 +153,36 @@ The following principles guide the evolution and interpretation of ICRC-3 and an
 
 ## Interaction with Other Standards
 
-Each standard that adheres to `ICRC-3` MUST define the list of block schemas that it introduces.  
-Each block schema MUST:
+Any standard that builds on ICRC-3 and defines new block types MUST clearly and unambiguously describe how those block types are recorded and interpreted.
 
-1. Extend the [Generic Block Schema](#generic-block-schema).
-2. Specify the expected value of `btype`. This MUST be unique across all standards.
-3. Describe the minimal `tx` structure required to interpret the block and determine its effect on ledger state.
-4. Define a canonical mapping from the method’s call parameters to the `tx` fields, including only parameters that are explicitly present in the user’s request.
+### When Defining a New Block Type
+For each block type introduced by a standard, the specification MUST:
 
-When defining how calls map to `tx`:
-- The `tx` must encode the user’s intent as passed in the method call.
-- Optional parameters not provided by the user MUST NOT appear in the `tx`.
-- If a top-level `fee` is present, it represents the effective fee charged by the ledger.  
-  If `tx.fee` is present, it represents the fee value specified by the user in the call.  
-  If both are absent, the fee is `0`.
-- The fee payer MUST be inferable from the block type alone.
-- Additional non-semantic fields may be added without requiring a new block type.  
-  Any new semantic fields that affect ledger state require a new block type definition.
+1. **Extend the [Generic Block Schema](#generic-block-schema)**.
+2. **Assign a unique `btype` value** that does not collide with any existing standard.
+3. **Describe the minimal `tx` structure** required to interpret the block and determine its effect on ledger state.
+4. **Specify the fee payer** for this block type.  
+   - The payer MUST be derivable from the block’s own fields (no external context or method-specific exceptions).
+   - This ensures the payer can be determined unambiguously by any ICRC-3-aware client.
+5. **Reference the [ICRC-3 effective fee rule](#principles-and-rules-for-icrc-3-blocks)** and state how it applies to this block type.  
+   - Clarify whether `tx.fee` is expected in user calls, when a top-level `fee` is present, and under what conditions either may be omitted.
+6. **Identify optional non-semantic fields** that may be added to `tx` without requiring a new block type.  
+   - Any change to the minimal semantic structure that affects ledger state requires defining a new `btype`.
+
+### When Defining a Canonical Mapping from Methods to Blocks
+For each method that produces a block:
+
+1. **Identify the `btype`** the method produces.
+2. **Define the canonical mapping** from the method’s call parameters to the `tx` fields of the resulting block:
+   - Include only parameters explicitly provided by the caller.
+   - Optional parameters not provided MUST NOT appear in `tx`.
+3. **Record fees according to the ICRC-3 fee rules**:
+   - `tx.fee` records the fee amount provided by the caller in the request.
+   - A top-level `fee` records the actual fee charged by the ledger.
+   - If both are absent, the fee is `0`.
+4. **Ensure operation names are namespaced** using the standard’s ICRC number as a prefix (see [Namespacing for Operations](#namespacing-for-operations)).
+
+This separation ensures that block semantics and method mappings are both specified with clarity and without cross-contamination.
 
 ### Namespacing for Operations
 
@@ -191,7 +204,9 @@ An ICRC-3 compatible Ledger MUST expose an endpoint listing all the supported bl
 
 ## [ICRC-1](../ICRC-1/README.md) and [ICRC-2](../ICRC-2/README.md) Block Schema
 
-This section describes how ICRC-1 and ICRC-2 operations are recorded in ICRC-3-compliant blocks.
+
+This section describes how ICRC-1 and ICRC-2 operations are represented in ICRC-3-compliant blocks.  These blocks follow the **legacy format**, meaning they do not have a `btype` field.  
+Instead, their type is inferred directly from the content of the `tx` field, which records the canonical mapping of the original method call.
 
 
 
--- a/standards/ICRC-3/README.md
+++ b/standards/ICRC-3/README.md
@@ -50,7 +50,7 @@ type Value = variant {
 };
 ```
 
-Servers must serve the block log as a list of `Value` where each `Value` represent a single block in the block log.
+Servers must serve the block log as a list of `Value` where each `Value` represents a single block in the block log.
 
 ## Value Hash
 
@@ -66,15 +66,15 @@ The hash function is the [representation-independent hashing of structured data]
 - the hash of an `Array` is the hash of the concatenation of the hashes of all the elements of the array
 - the hash of a `Map` is the hash of the concatenation of all the hashed items of the map sorted lexicographically. A hashed item is the tuple composed by the hash of the key and the hash of the value.
 
-Pseudocode for representation independent hashing of Value, together with test vectors to check compliance with the specification can be found [`here`](HASHINGVALUES.md). 
+Pseudocode for representation-independent hashing of `Value`, together with test vectors to check compliance with the specification can be found [`here`](HASHINGVALUES.md). 
 
 ## Blocks Verification
 
-The Ledger MUST certify the last block (tip) recorded. The Ledger MUST allow to download the certificate via the `icrc3_get_tip_certificate` endpoint. The certificate follows the [IC Specification for Certificates](https://internetcomputer.org/docs/current/references/ic-interface-spec#certification). The certificate is comprised of a tree containing the certified data and the signature. The tree MUST contain two labelled values (leafs):
-1. `last_block_index`: the index of the last block in the chain. The values must be expressed as [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128)
+The Ledger MUST certify the last block (tip) recorded. The Ledger MUST allow to download the certificate via the `icrc3_get_tip_certificate` endpoint. The certificate follows the [IC Specification for Certificates](https://internetcomputer.org/docs/current/references/ic-interface-spec#certification). The certificate is comprised of a tree containing the certified data and the signature. The tree MUST contain two labeled values (leaves):
+1. `last_block_index`: the index of the last block in the chain. The values MUST be expressed as [`leb128`](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128)
 2. `last_block_hash`: the hash of the last block in the chain
 
-Clients SHOULD download the tip certificate first and then download the block backward starting from `last_block_index` and validate the blocks in the process.
+Clients SHOULD download the tip certificate first and then download the blocks backward starting from `last_block_index` and validate the blocks in the process.
 
 Validation of block `i` is done by checking the block hash against
 1. if `i + 1 < len(chain)` then the parent hash `phash` of the block `i+1`
@@ -86,7 +86,7 @@ An ICRC-3 compliant Block
 
 1. MUST be a `Value` of variant `Map`
 2. MUST contain a field `phash: Blob` which is the hash of its parent if it has a parent block
-3. SHOULD contain a field `btype: String` which uniquely describes the type of the Block. If this field is not set then the block type falls back to ICRC-1 and ICRC-2 for backward compatibility purposes
+3. SHOULD contain a field `btype: Text` which uniquely describes the type of the Block. If this field is not set then the block type falls back to ICRC-1 and ICRC-2 for backward compatibility purposes
 
 
 ### Kinds of Blocks
--- a/standards/ICRC-3/README.md
+++ b/standards/ICRC-3/README.md
@@ -245,7 +245,7 @@ A legacy block:
 
 - **MUST** be a `Value::Map` containing at least:
   - `"phash"`: `Blob` — the parent hash.
-  - `"ts"`: `Nat` — the timestamp set by the ledger when the block was created.
+  - `"ts"`: `Nat` — the timestamp (in nanoseconds since Unix epoch) set by the ledger when the block was created.
   - `"tx"`: `Value::Map` — representing the user’s transaction intent.
 - **MAY** include:
   - `"fee": Nat` — the fee actually charged by the ledger, if any.
--- a/standards/ICRC-3/README.md
+++ b/standards/ICRC-3/README.md
@@ -438,41 +438,6 @@ icrc1_transfer: record {
 - `to` and `fee` MUST NOT be present
 
 
-### Fee Payer and Balance Effects (Legacy ICRC-1/2)
-
-The rules below define **who pays** and how the **effective fee** (if any) affects balances for legacy blocks (no `btype`; kind inferred from `tx`). The authoritative charged amount is the top-level `fee : Nat` when present; `tx.fee` (if present) reflects the caller input only.
-
-#### `icrc1_transfer`
-- `op = "xfer"` → **Payer:** `from`  
-  • Debited from `from`: `amt + fee` (if a fee is charged)  
-  • Credited to `to`: `amt`
-- `op = "burn"` → **Payer:** `from`  
-  • Debited from `from`: `amt + fee` (if a fee is charged)  
-  • Burned: `amt`
-- `op = "mint"` → **Payer:** `to`  
-  • Credited to `to`: `amt - fee` (if a fee is charged; require `fee ≤ amt`)  
-  • Minted gross amount: `amt` (with `fee` immediately taken from `to`)
-
-#### `icrc2_transfer_from`
-- `op = "xfer"` → **Payer:** `from` (authorized by `spender`)  
-  • Debited from `from`: `amt + fee` (if a fee is charged)  
-  • Credited to `to`: `amt`
-- `op = "burn"` → **Payer:** `from` (authorized by `spender`)  
-  • Debited from `from`: `amt + fee` (if a fee is charged)  
-  • Burned: `amt`
-- `op = "mint"` → Payer: to (authorized by spender under an approval on the minting account)
-• Credited to `to: amt - fee` (if a fee is charged; require `fee ≤ amt`)
-• Allowance on the minting account (for spender) is reduced by `amt + fee`
-
-#### `icrc2_approve`
-- **Payer:** `from` (the account whose allowance is modified)  
-  • Debited from `from`: `fee` (if a fee is charged)
-
-**Notes**  
-- A fee may be charged even if `tx.fee` is absent; the charged fee is indicated by a top-level `fee`.  
-- If no top-level `fee` is present, the effective fee is `0`.  
-- Implementations must reject calls that cannot satisfy the fee rule (e.g., `fee > amt` for mint; or insufficient balance for `amt + fee` debits).
-
 
 
 ### Canonical Examples of `icrc1_transfer` Blocks
--- a/standards/ICRC-3/README.md
+++ b/standards/ICRC-3/README.md
@@ -50,13 +50,13 @@ type Value = variant {
 };
 ```
 
-Servers must serve the block log as a list of `Value` where each `Value` represents a single block in the block log.
+Servers MUST serve the block log as a list of `Value` where each `Value` represents a single block in the block log.
 
 ## Value Hash
 
 `ICRC-3` specifies a standard hash function over `Value`.
 
-This hash function should be used by Ledgers to calculate the hash of the parent of a block and by clients to verify the downloaded block log.
+This hash function SHOULD be used by Ledgers to calculate the hash of the parent of a block and by clients to verify the downloaded block log.
 
 The hash function is the [representation-independent hashing of structured data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map) used by the IC:
 - the hash of a `Blob` is the hash of the bytes themselves
@@ -109,36 +109,36 @@ Block types and their schemas are defined either by legacy standards (e.g., ICRC
 The following principles guide the evolution and interpretation of ICRC-3 and any standards that build on it.
 
 ### 1. Core State Transitions
-- Every block type must define the **core state transition** it represents: the deterministic change to ledger state implied by the block’s minimal `tx` structure, *ignoring fees or ledger-specific policies*.  
+- Every block type MUST define the **core state transition** it represents: the deterministic change to ledger state implied by the block’s minimal `tx` structure, *ignoring fees or ledger-specific policies*.  
 - This transition is the canonical meaning of a block — what balances, allowances, or other state variables change as a direct consequence of the block.  
 - Fee handling, metadata, and ledger-specific policies are layered on top of this transition.
 
 ### 2. Separation of `btype` and `tx`
 - The `btype` field defines the **minimal semantic structure** of a block — the set of fields in `tx` required to fully determine its core state transition.  
-- Standards that introduce a new `btype` must:
+- Standards that introduce a new `btype` MUST:
   - Assign a unique identifier for the `btype`.
   - Specify the minimal `tx` structure required for interpreting that block type.
   - Define the block’s **core state transition** in terms of this minimal structure.
-- Standards that define methods producing blocks must:
+- Standards that define methods producing blocks MUST:
   - Specify which `btype` the method produces.
   - Define the **canonical mapping** from method call parameters to the `tx` field of the resulting block.
   - Ensure that `tx` contains only parameters explicitly provided by the caller (except where the block type definition requires otherwise).
 
 ### 3. Avoiding Collisions in `tx`
-- To avoid collisions between transactions originating from different standards, the canonical `tx` mapping must include:
+- To avoid collisions between transactions originating from different standards, the canonical `tx` mapping MUST include:
   - An operation field (`op`) whose value is namespaced using the standard’s number as a prefix, e.g., `122freeze_account`.
 - No two standardized methods may produce `tx` values that are indistinguishable when interpreted under ICRC-3 rules.
 
 ### 4. Inclusion of the User Call in `tx`
 - The `tx` field must faithfully capture the structure of the user call that triggered the block.
-- All call parameters that are part of the method’s canonical mapping must be included exactly as provided by the caller.
-- Optional parameters that were not present in the call must be omitted from `tx`.
+- All call parameters that are part of the method’s canonical mapping MUST be included exactly as provided by the caller.
+- Optional parameters that were not present in the call MUST be omitted from `tx`.
 
 ### 5. Future-Proofing and Extensibility
-- Additional non-semantic fields (e.g., metadata, hashes, references) may be added to `tx` without introducing a new `btype`, provided:
+- Additional non-semantic fields (e.g., metadata, hashes, references) MAY be added to `tx` without introducing a new `btype`, provided:
   - They do not affect the block’s **core state transition**.
   - They are ignored by block verification and interpretation logic that only relies on the minimal `tx` structure defined by the `btype`.
-- Any change to the minimal semantic structure of a block requires introducing a new `btype`.
+- Any change to the minimal semantic structure of a block REQUIRES introducing a new `btype`.
 
 ### Note on Ledger-Specific Fields
 - Blocks may include additional fields specific to a given standard or ledger (e.g., `fee`, metadata, references).  
@@ -162,7 +162,7 @@ To ensure consistency across standards and implementations, the semantics of any
    • Example: debit/credit balances, mint, burn, update allowance.
 
 4. Apply fee (if applicable)  
-   • If the block type involves fees, determine the **effective fee** following ICRC-107 rules.  
+   • If the block type involves fees, determine the **effective fee** following ICRC-107.  
    • Deduct the fee from the account designated as the **fee payer** for this block type.  
    • Adjust balances accordingly (e.g., for mints: `to` receives `amt – fee`).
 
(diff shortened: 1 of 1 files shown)


standards/ICRC-3/README.md
```diff
@@ -88,236 +88,640 @@ An ICRC-3 compliant Block
...
+**Structure**
+- **MUST** contain `tx.op = "mint"`.
+- **MUST** contain `tx.to : Account`.
+- **MUST** contain `tx.amt : Nat`.
+- **MUST NOT** contain `tx.from`.
+- **MAY** contain `tx.memo : Blob` if provided by the caller.
+- **MAY** contain `tx.ts : Nat` if provided by the caller.
+
+**Semantics**  
+Mints create `amt` new tokens. If a fee is charged, it is deducted from `to` immediately, so `to` receives `amt - fee` (require `fee ≤ amt`).
```
mbjorkqvist, line 286: I find this to be inconsistent with `transfer` and `burn` - for the other operations, the total amount is `amt + fee`, whereas here with `mint`, the total is `amt`. On the other hand, if `to` receives `amt`, then `amt + fee` is minted, and `Fee payer` would be `None`.

standards/ICRC-3/README.md
```diff
@@ -88,236 +88,640 @@ An ICRC-3 compliant Block
...
+  • Credited to `to`: `amt - fee` (if a fee is charged; require `fee ≤ amt`)  
+  • Minted gross amount: `amt` (with `fee` immediately taken from `to`)
+
+#### `icrc2_transfer_from`
+- `op = "xfer"` → **Payer:** `from` (authorized by `spender`)  
+  • Debited from `from`: `amt + fee` (if a fee is charged)  
+  • Credited to `to`: `amt`
+- `op = "burn"` → **Payer:** `from` (authorized by `spender`)  
+  • Debited from `from`: `amt + fee` (if a fee is charged)  
+  • Burned: `amt`
```
mbjorkqvist, line 456: Should `op = "mint"` also be included for `icrc2_transfer_from`? The DFINITY ICRC ledger implementation does not allow delegated mints, but it doesn't seem to be prohibited in the ICRC-2 standard.

standards/ICRC-3/README.md
```diff
@@ -88,236 +88,640 @@ An ICRC-3 compliant Block
...
+  - An operation field (`op`) whose value is namespaced using the standard’s number as a prefix, e.g., `122freeze_account`.
+- No two standardized methods may produce `tx` values that are indistinguishable when interpreted under ICRC-3 rules.
+
+### 4. Inclusion of the User Call in `tx`
+- The `tx` field must faithfully capture the structure of the user call that triggered the block.
+- All call parameters that are part of the method’s canonical mapping must be included exactly as provided by the caller.
+- Optional parameters that were not present in the call must be omitted from `tx`.
+
+### 5. Future-Proofing and Extensibility
+- Additional non-semantic fields (e.g., metadata, hashes, references) may be added to `tx` without introducing a new `btype`, provided:
```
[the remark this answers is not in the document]
bogwar: To the standard, since adding fields to an implementation that are not documented here would not allow one to recover tx. 
This would still run into the issue of older clients who would not keep up with the standard. 
Perhaps it's best to just not comment on extensibility since it brings more questions that it solves.

standards/ICRC-3/README.md
bogwar: Good question: this would allow for some nice usecases where one can delegate bounded minting to third parties. 
Let's include this in our discussion.

standards/ICRC-3/README.md
bogwar: After discussion, we agreed that:
* it would require to allow for blocks where the fee payer is not specified, in which case the fee still exists (is minted) but it is not deducted from anywhere.
*  minting more than the specified amount may have undesirable repercussions for stable coins and RWA backed ledgers, if not careful. 

To eliminate the footgun we propose to leave the semantics as is.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,642 @@ An ICRC-3 compliant Block
...
+
+**Transfer to the Minting Account (→ Burn)** — when `to` equals the minting account:
+
+- `op = "burn"`
+- `from = [caller]` if `from_subaccount` is not provided  
+- `from = [caller, from_subaccount]` if provided
+- `amt = amount`
+- `memo = memo` if provided
+- `ts = created_at_time` if provided  
+- `to` and `fee` MUST NOT be present
```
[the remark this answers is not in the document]
marc0olo: so we cannot allow `burn` blocks to have a fee for icrc1, right? can you elaborate shortly again why we would want to add a burn fee? 🤔

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,661 @@ An ICRC-3 compliant Block
...
+          record { "op"; variant { Text = "burn" } };
+        }
+      };
+    };
+  }
+};
 ```
 
-Example without `btype`:
+##### `icrc2_transfer_from`
```
andreacerulli, line 592: Should this be a subsection of `Canonical tx Mapping` I am a bit lost in the document.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,661 @@ An ICRC-3 compliant Block
...
+**Transfer from the Minting Account (→ Mint)** — when `[caller]` or `[caller, from_subaccount]` equals the minting account:
+
+- `op = "mint"`
+- `to = to`
+- `amt = amount`
+- `memo = memo` if provided
+- `ts = created_at_time` if provided  
+- `from` and `fee` MUST NOT be present
+
+**Transfer to the Minting Account (→ Burn)** — when `to` equals the minting account:
```
andreacerulli, line 430: What about when both `caller` and `to` are the minting account? Similarly also for transfer from

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,642 @@ An ICRC-3 compliant Block
...
+  - Define the block’s **core state transition** in terms of this minimal structure.
+- Standards that define methods producing blocks MUST:
+  - Specify which `btype` the method produces.
+  - Define the **canonical mapping** from method call parameters to the `tx` field of the resulting block.
+  - Ensure that `tx` contains only parameters explicitly provided by the caller (except where the block type definition requires otherwise).
+
+### 3. Avoiding Collisions in `tx`
+- To avoid collisions between transactions originating from different standards, the canonical `tx` mapping MUST include:
+  - An operation field (`op`) whose value is namespaced using the standard’s number as a prefix, e.g., `122freeze_account`.
+- No two standardized methods may produce `tx` values that are indistinguishable when interpreted under ICRC-3 rules.
```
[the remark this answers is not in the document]
bogwar: Equal 'tx' would have the same 'tx.op'. 

This bullet point deals with the following issue:
The first requirement is not tight enough to ensure that this doesn't happen (since it only demands that the standard number appears in the op, not that the op is tied somehow to the method call).  Perhaps something better would be to require that out of tx one can recover the original call, unambigously.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,661 @@ An ICRC-3 compliant Block
...
+- **MUST** contain `tx.to : Account`.
+- **MUST** contain `tx.amt : Nat`.
+- **MUST NOT** contain `tx.from`.
+- **MAY** contain `tx.spender : Account`.
+- **MAY** contain `tx.memo : Blob` if provided by the caller.
+- **MAY** contain `tx.ts : Nat` if provided by the caller.
+
+**Semantics**  
+Mints create `amt` new tokens. If a fee is charged, it is deducted from `to` immediately, so `to` receives `amt - fee` (require `fee ≤ amt`).  
+If `tx.spender` is present, the mint is executed under an approval on the minting account; that approval **MUST** be at least `amt + fee` and **MUST** be reduced by `amt + fee`.
```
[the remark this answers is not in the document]
bogwar: Good catch! Thanks.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,642 @@ An ICRC-3 compliant Block
...
+  - Specify the minimal `tx` structure required for interpreting that block type.
+  - Define the block’s **core state transition** in terms of this minimal structure.
+- Standards that define methods producing blocks MUST:
+  - Specify which `btype` the method produces.
+  - Define the **canonical mapping** from method call parameters to the `tx` field of the resulting block.
+  - Ensure that `tx` contains only parameters explicitly provided by the caller (except where the block type definition requires otherwise).
+
+### 3. Avoiding Collisions in `tx`
+- To avoid collisions between transactions originating from different standards, the canonical `tx` mapping MUST include:
+  - An operation field (`op`) whose value is namespaced using the standard’s number as a prefix, e.g., `122freeze_account`.
```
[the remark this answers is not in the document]
bogwar: Not quite: collisions in tx need to be avoided across standards (irrespective of the block types that are created) so that one can uniquely identify the transaction. The proposal is that the canonical mapping of calls to tx include the number of the standard where the method is defined. 
This will guarantee that methods defined in different standards will not lead to identical tx. 

For example (under these rules), 1xfer and 2xfer would be the ops that correspond to icrc1 transfer and icrc2 transfer_from.  

The point below should ensure that methods defined within the same standard do not lead to colliding tx's

standards/ICRC-3/README.md
bogwar: There's no good answer here, and although there are some restrictions in ICRC-2, this case is really underspecified. 
I suggest to simply prevent it explicitly. WDYT?

standards/ICRC-3/README.md
bogwar: Yes, it became that after removing the section above.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
+ICRC-3 defines how blocks are structured and verified. Other standards extend this by either:  
+(1) introducing new block types (`btype`), or  
+(2) defining canonical mappings from standardized method calls to existing block types.
+
+### Standards That Introduce Block Types
+A standard that defines a new block type MUST:
+- Assign a unique `btype`.  
+- Specify the minimal `tx` structure required to interpret the block and determine its effect on ledger state.  
+- Define semantics using the **Semantics of Blocks: Evaluation Model** (pre-fee transition, fee hook, post-conditions).  
+- If the block type involves fees, reference the applicable fee standard (e.g., ICRC-107) and **define who pays**, via a fee payer expression resolvable from block fields.
```
maciejdfinity: If I understand section `4. Apply fee (if applicable)` above correctly, `ICRC-107` is not optional?

maciejdfinity: Now I read the `Note on Fees` and it says that standards defining fee handling should follow `ICRC-107`. Does it mean that standards that introduce block types can use a standard that follows `ICRC-107` but is possibly distinct from `ICRC-107` and it should always be possible to determine the effective fee by only following `ICRC-107`?

maciejdfinity: `Note on Fees` also says that 107 MUST be followed.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
+- **MUST** contain `tx.op = "burn"`.
+- **MUST** contain `tx.from : Account`.
+- **MUST** contain `tx.amt : Nat`.
+- **MUST NOT** contain `tx.to`.
+- **MAY** contain `tx.fee : Nat` if provided by the caller.
+- **MAY** contain `tx.memo : Blob` if provided by the caller.
+- **MAY** contain `tx.ts : Nat` if provided by the caller.
+
+**Semantics**  
+Burns remove `amt` tokens from `from`. Any fee is also debited from `from`.
```
maciejdfinity:
```suggestion
Burns remove `amt + fee` tokens from `from`.  
```

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
+---
+
+#### Burn Block (`op = "burn"`)
+
+**Structure**
+- **MUST** contain `tx.op = "burn"`.
+- **MUST** contain `tx.from : Account`.
+- **MUST** contain `tx.amt : Nat`.
+- **MUST NOT** contain `tx.to`.
+- **MAY** contain `tx.fee : Nat` if provided by the caller.
```
maciejdfinity: Do I understand correctly that burn transactions may specify the fee and mint transactions cannot? I.e. for a mint transaction to have fee, it has to be specified at the block level?

maciejdfinity: Below in the `Transfer to the Minting Account (→ Burn)` section it says the  `fee` MUST NOT be present.

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
-4. it CAN contain a field `tx.expected_allowance: Nat` if set by the user
-5. it CAN contain a field `tx.expires_at: Nat` if set by the user
+**Burn Transfer** — when the `to` account is the minting account:
+
+- `op = "burn"`
+- `from = from` (as passed in the call)
+- `spender = [caller]` if `spender_subaccount` is not provided
+- `spender = [caller, spender_subaccount]` if provided
+- `amt = amount`
+- `fee = fee` if provided
```
maciejdfinity: Above in the `Transfer to the Minting Account (→ Burn)` section it says the `fee` MUST NOT be present.

standards/ICRC-3/README.md
bogwar: There are potentially two issues: 
* 103 precedes 107 so why should it reference a standard the follows it
* for ledgers that do not report compliance with 107, the fee behavior is then underspecified. 

As I see it:
* Although 107 is not yet out, it should be by the time this is merged so it should be ok to reference it. 
* Not reporting compliance with 107 doesn't leave to a worse situation than the current one where the fee behavior was a bit ad-hoc. So we can only improve on the situation by declaring (and implementing) compliance with 107

standards/ICRC-3/README.md
bogwar: `fee` is not fully defined here: it doesn't refer to `tx.fee` but to the effective fee of a block. 
Perhaps it is worth explaining that.

standards/ICRC-3/README.md
bogwar: Right -- if the user specifies a fee then the call should actually be rejected (since otherwise we'd have to record the fee, which would mess up with our specification of `effective_fee`.

standards/ICRC-3/README.md
maciejdfinity: I am merely pointing out inconsistencies between the 3 sections that mention ICRC-107. Two of them state that it is mandatory, one that it is optional.

standards/ICRC-3/README.md
maciejdfinity: So this line should be removed?

standards/ICRC-3/README.md
bogwar: Following 1:1 -- prevent these cases below in the tx mapping for ICRC1/2 methods

standards/ICRC-3/README.md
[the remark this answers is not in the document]
bogwar: I'll switch these around

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
-        record { "amt"; variant { Nat = 1_228_990 : nat } };
-        record { "from"; variant { Array = vec {
-                variant { Blob = blob "\00\00\00\00\020\00\07\01\01" };
-                variant { Blob = blob "&\99\c0H\7f\a4\a5Q\af\c7\f4;\d9\e9\ca\e5 \e3\94\84\b5c\b6\97/\00\e6\a0\e9\d3p\1a" };
-        }}};
-        record { "memo"; variant { Blob = blob "\82\00\83x\223K7Bg3LUkiXZ5hatPT1b9h3XxJ89DYSU2e\19\07\d0\00"
+variant { Array = vec {
+                variant { Blob = blob "\00\00\00\00\00\f0\13x\01\01" };
+                variant { Blob = blob "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00" };
         }};
```
q-uint: NIT: reformat

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,626 @@ An ICRC-3 compliant Block
...
-        record { "amt"; variant { Nat = 1_228_990 : nat } };
-        record { "from"; variant { Array = vec {
-                variant { Blob = blob "\00\00\00\00\020\00\07\01\01" };
-                variant { Blob = blob "&\99\c0H\7f\a4\a5Q\af\c7\f4;\d9\e9\ca\e5 \e3\94\84\b5c\b6\97/\00\e6\a0\e9\d3p\1a" };
-        }}};
-        record { "memo"; variant { Blob = blob "\82\00\83x\223K7Bg3LUkiXZ5hatPT1b9h3XxJ89DYSU2e\19\07\d0\00"
+variant { Array = vec {
+                variant { Blob = blob "\00\00\00\00\00\f0\13x\01\01" };
+            
         }};
```
q-uint: NIT: reformat

q-uint (commented):
We're currently writing some (internal) libraries for ICRC3, would be great to have this merged!

Request: would be great to have a test suite which can just be a list of valid blocks.
I.e. `burn`, `1burn`, `mint`, `1mint`, etc. (maybe separated by generic/legacy).

Thanks for the work!

standards/ICRC-3/README.md
bogwar: good catch

standards/ICRC-3/README.md
bogwar: Removed it

standards/ICRC-3/README.md
```diff
@@ -86,238 +86,661 @@ An ICRC-3 compliant Block
...
+- **MAY** contain `tx.fee : Nat` if provided by the caller.
+- **MAY** contain `tx.memo : Blob` if provided by the caller.
+- **MAY** contain `tx.ts : Nat` if provided by the caller (`created_at_time`).
+- **MAY** contain `tx.spender : Account` if created via `icrc2_transfer_from`.
+
+**Semantics**  
+Transfers debit `amt` (and any fee) from `from` and credit `amt` to `to`.  
+If `tx.spender` is present, the operation is executed under an approval, which must cover at least `amt + fee`. The allowance is reduced accordingly.  
+
+**Fee payer:** `from`.
```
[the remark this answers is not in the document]
bogwar: moved to tx.x everywhere

standards/ICRC-3/README.md
bogwar: added a note spelling out that that is not allowed

bogwar:
> We're currently writing some (internal) libraries for ICRC3, would be great to have this merged!
> 
> Request: would be great to have a test suite which can just be a list of valid blocks. I.e. `burn`, `1burn`, `mint`, `1mint`, etc. (maybe separated by generic/legacy).
> 
> Thanks for the work!

Thanks for the comments!

Could you be a bit more specific about the form of such a test suite? 

According to the current version of the standard there should be no `1burn` btype, but rather this type would need to be inferred from the form of the generic blocks.

standards/ICRC-3/README.md
bogwar: Adjusted. Please check.

standards/ICRC-3/README.md
bogwar: not sure what the reformatting refers to. I've adjusted things a bit. Please check.

q-uint:
Hi @bogwar, sorry the `1burn` was a typo.

It seems that currently we are going to need to support both legacy and the new block log types. As this gets somewhat confusing, it would be nice to have a test suite with examples of each of these `Value` block types.

Personally I don't need these anymore, but it was a lot of work to construct tests which cover all possible flavors of block types. In the end we just ran the indexer against an SNS ledger to battle test it.

standards/ICRC-3/README.md
q-uint: I was just referring to the indentation, it has been fixed! ✅

q-uint:
Thanks for your time and work on this @bogwar! We're working toward supporting ICRC3 on non-ledger canisters, and your updates make the event log design clearer and more suitable for generic use cases.

q-uint:
I don't want to block this PR from being merged but while applying ICRC-3 to our internal event log I encountered some 'blockers'.

### `Value` type
- `tuple` support?
- `bool` support?
- `null` support?
- (Optional) `principal` support?

This is how I currently circumvent this:

```rust
impl ICRC3Value for Principal {
    fn to_value(&self) -> Value {
        // This is ok, since principal is a "blob".
        Value::Blob(self.as_slice().to_vec().into())
    }
}

impl ICRC3Value for bool {
    fn to_value(&self) -> Value {
        Value::Nat(Nat::from(*self as u8))
    }
}
```

Currently we can only support optional values in `Map` since this allows you to leave them out, yet we can not do the same in, i.e., `Array`, to represent tuples.

bogwar:
Thanks Quint!

On extending Value:
Value was intentionally minimal (the design was entirely ledger-driven). I’d prefer to keep it that way unless there’s a strong, recurring need.

Principals / bools:
Encoding principals as Blob is fine; bool as Nat 0/1 is also reasonable.

Tuples with optionals:
The tricky case is tuples containing opt. Two workable patterns:

- Prefer Maps over tuples when fields can be optional (omit absent fields).

- If you must keep array/tuple semantics, use a tagged option per slot: None = [Nat 0], Some(x) = [Nat 1, x].

I’d suggest we keep Value unchanged and add a non-normative “Common Encodings” section with these recommendations (principal→Blob, bool→Nat 0/1, options in tuples via tags). That preserves minimalism while giving interoperable guidance. WDYT?

bogwar:
@q-uint Hi Quint, can you please take a look? I've added a non-normative section with encodings.

standards/ICRC-3/README.md
```diff
@@ -446,7 +450,11 @@ Although legacy ICRC-1 and ICRC-2 blocks do not include the `btype` field, ledge
 
 ### Account Type
```
letmejustputthishere, line 949: i still think it would be valuable to have a `Principal` variant in the `Value`. this way there is no ambiguity on how to interpret arrays

standards/ICRC-3/README.md
```diff
@@ -50,272 +112,773 @@ type Value = variant {
 };
 ```
 
-Servers must serve the block log as a list of `Value` where each `Value` represent a single block in the block log.
+Servers MUST serve the block log as a list of `Value` where each `Value` represents a single block in the block log.
+
+## Common Encodings (Non-Normative)
```
letmejustputthishere, line 117: i still think it would be valuable to extend `Value` to avoid ambiguity when decoding logs.

sea-snake: Ambiguity should be avoidable due to the requirement of block types explicitly specifying their expected data structure.

Adding more types to Value was considered previously, but we didn't opt for it due to:
- It's a significantly larger breaking change to ICRC-3, existing tooling would not know how to handle the additional Value types.
- How many different types do we add to Value? The original intent of Value was to have a bare minimum data structure. If we add all candid types to it, it only becomes an extra wrapper around candid, using the encoded candid bytes directly instead starts to make more sense at that point.
- As mentioned above, ambiguity should be solely avoided by clearly defining block schemas.

standards/ICRC-3/README.md
gregorydemay, line 74: Should the diagram maybe mention the other mandatory fields (like `ts`)?

standards/ICRC-3/README.md
```diff
@@ -50,293 +112,266 @@ type Value = variant {
...
-    Map : vec record { text; Value };
-};
+- Blocks produced by that method SHOULD include a `tx` field.
+- The `tx` field for such blocks SHOULD include an operation discriminator `op`.
+- The `op` value SHOULD be namespaced with the standard’s ICRC number  
+  (e.g., `122freeze_account`, `107set_fee_collector`) to avoid collisions with operations defined by other standards.
+
+Typed blocks that **do not** represent method calls (e.g., upgrade markers, maintenance events, migration records, or system actions) MAY omit the `tx` field entirely and therefore MAY omit `tx.op`.
+
+Legacy ICRC-1/2 blocks continue to use their historical operation names (`"xfer"`, `"mint"`, `"burn"`, `"approve"`), and are exempt from namespacing requirements.
```
gregorydemay, line 268: I don't understand this part. Those names come from **our** implementation of the ICRC ledgers, but AFAICT they are not mentioned anywhere in the [ICRC-1](https://github.com/dfinity/ICRC-1/blob/46e77a035f0683ee820f89e9b771b72b0af43918/standards/ICRC-1/README.md) or [ICRC-2](https://github.com/dfinity/ICRC-1/blob/46e77a035f0683ee820f89e9b771b72b0af43918/standards/ICRC-2/README.md) standards? In fact ICRC-1 only mentions the notion of block as being [excluded](https://github.com/dfinity/ICRC-1/blob/46e77a035f0683ee820f89e9b771b72b0af43918/standards/ICRC-1/README.md?plain=1#L158), while ICRC-2 does not talk about blocks at all.

standards/ICRC-3/README.md
```diff
@@ -50,293 +112,266 @@ type Value = variant {
...
+
+
+### 4. Capturing the User Call (Where Applicable)
+
+If a block represents the result of a **standardized user-initiated method call**, then:
+
+- The block **SHOULD** include a `tx` field.
+- For user-initiated blocks, a namespaced `tx.op` **SHOULD** be included (see above).
```
gregorydemay: Understanding question: what's the difference between`tx.op` and `btype`? Is-it because a `btype` block could be produced by several user endpoints?

standards/ICRC-3/README.md
```diff
@@ -50,293 +112,266 @@ type Value = variant {
...
+   • Semantic fields may appear **inside `tx`** or **at the top level**, depending on the
+     block-type specification.  
+   • Additional non-semantic fields MAY appear anywhere in the block and MUST be
+     ignored by generic interpreters.  
+   • ICRC-3 itself does not distinguish between “semantic” and “non-semantic”
+     fields; it is the responsibility of the block type specification to state
+     which fields affect the meaning of the block and how extra fields are to be
+     treated (e.g., ignored by generic interpreters).
+
+3. Derive pre-fee state transition
```
gregorydemay, line 330: I don't really understand this part (also apply fee below), which seems ledger specific (it talks about fees) but that whole section (` Semantics of Blocks: Evaluation Model`) should be domain-agnostic way, no?

standards/ICRC-3/README.md
```diff
@@ -351,30 +386,113 @@ type GetArchivesResult = vec record {
...
+
+- For every archive returned:
+  - `start` and `end` MUST describe a contiguous, inclusive block range;
+  - that archive MUST be able to serve exactly that range via its callbacks (as returned indirectly in `icrc3_get_blocks`).
+- The producer MAY change its archival layout over time (e.g., due to upgrades, reallocation, compaction, or   reinstallation). The values of `canister_id`, `start`, and `end` MUST correctly reflect the archival layout at the time of the call.
+- A producer MUST NOT return overlapping or out-of-order ranges across the union of all archives.
+
+### `icrc3_get_blocks`
+Returns blocks for one or more requested ranges from the canonical block log.
+Blocks that are no longer stored locally by the producer are returned indirectly via archive callbacks.
```
gregorydemay, line 410: is-it correct that the callback is not-necessarily an ICRC-3 compliant method?

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+
+- If the producer canister has not yet emitted any blocks, it MAY return `null`.
+
+- Otherwise, the method MUST return an IC data certificate whose `hash_tree` authenticates a labeled subtree containing at least:
+  - `last_block_index` — the leb128-encoded index of the last block, and
+  - `last_block_hash` — the hash of that block.
+
+- Additional labeled values MAY be present in the certified data tree.
+
+- The certificate MUST be a valid IC data certificate as defined in the IC interface specification. In particular, its signature and delegation MUST verify against the IC root key, and the `hash_tree` MUST be consistent with the certified data used to derive `last_block_index` and `last_block_hash` at the time the certificate was created.
```
gregorydemay, line 523: nit: Should we put the IC root public key in the standard?

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+- **Specify the minimal structure** required to interpret the block and recover its
+  semantic meaning.  
+  This structure MUST contain all fields needed to reconstruct the event’s
+  effect. These semantic fields may live inside `tx`, at the top level, or be
+  split between the two, as explicitly defined by the block-type specification.
+
+- **Define precise semantics** for the block using the  
+  *Semantics of Blocks: Evaluation Model*:  
+  - pre-fee state transition,  
+  - fee payer (if applicable),
```
gregorydemay, line 587: This seems ledger specific

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+
+- **Describe how fees are handled**, if the block type involves fees:  
+  - The standard **SHOULD** specify the **effective fee** (the fee that is
+    actually charged), where this is well-defined.  
+  - The standard **SHOULD** specify the **fee payer**, as an expression
+    resolvable from fields in the block, where this is possible.  
+  - If applicable, the standard **MAY** reference ICRC-107 to specify **where
+    the fee goes** (e.g., burned, sent to a collector, etc.).  
+  - If some fee-related aspects are intentionally left implementation- or
+    ledger-defined, the standard SHOULD say so explicitly.
```
gregorydemay, line 599: This seems ledger specific

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+- Every standard that introduces a block type involving fees **SHOULD** specify who the
+  fee payer is and, where possible, how the effective fee is determined, so that
+  responsibility is unambiguous.  
+- Where this cannot be expressed generically (for example, because the policy is
+  intentionally ledger-defined), the standard SHOULD state that the fee payer and/or
+  fee amount are implementation- or ledger-specific.  
+- The rules for interpreting the amount and destination of fees MAY be delegated to
+  ICRC-107 (Fee Handling in Blocks). Ledgers that do not yet implement ICRC-107 MAY
+  still produce valid ICRC-3 blocks, but their fee behavior will remain
+  implementation-specific until aligned with ICRC-107.
```
gregorydemay, line 687: Seems ledger specific

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+
+
+
+
+## Supported Standards
+
+An ICRC-3 compatible producer canister MUST expose an endpoint listing all the supported block types via the endpoint `icrc3_supported_block_types`.
+
+- For **typed** blocks, the producer canister MUST only produce blocks whose `"btype"` value is included in this list.
+- For **legacy** ICRC-1/2 blocks (no `"btype"`), the producer canister MUST include the conventional identifiers (e.g., `"1xfer"`, `"2approve"`) in this list to advertise support, even though the blocks themselves do not carry a `"btype"` field.
```
gregorydemay, line 697: Similar comment as above, I don't see in which standard those are specified.

standards/ICRC-3/README.md
```diff
@@ -392,10 +510,823 @@ service : {
...
+  // Returns the set of block types this producer may emit
+  icrc3_supported_block_types : () -> (vec SupportedBlockType) query;
+}
+```
+
+
+
+
+
+## [ICRC-1](../ICRC-1/README.md) and [ICRC-2](../ICRC-2/README.md) Block Schema
```
gregorydemay, line 804: Understanding question: why are we trying to retrofit our ICRC-1 and ICRC-2 implementations into the new ICRC-3 standard? Why not just say that the ledger X is ICRC-3 compliant starting block `b` and verifying past blocks `< b` involves knowing the implementation (which is needed anyway and describe din this section but should not maybe be part of ICRC-3)

gregorydemay (commented):
Thanks @bogwar For revamping this! Some understanding questions and minor comments. Happy to discuss offline if that helps.

standards/ICRC-3/README.md
bogwar: The focus of the diagram is on the chaining aspect; I would keep it like this.

PR #202, opened 2025-09-08
chore: FI-1858: Bump crate versions and deprecate icrc1-test-env-state-machine

mbjorkqvist:
Bump the crate versions to `0.1.12` and mark the `icrc1-test-env-state-machine` as deprecated, since it will be migrated to/replaced by the `icrc1-test-env-pocket-ic` crate.

### DIFF
--- a/test/env/Changelog.md
+++ b/test/env/Changelog.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.12] - 2025-09-08
+### Changed
+- Bump rust to 1.85.0
+
+## [0.1.11] - 2024-01-16
+### Added
+- Duplicate release of 0.1.2
+
 ## [0.1.2] - 2024-01-16
 ### Changed
 - Use candid 0.10
--- a/test/env/state-machine/Cargo.toml
+++ b/test/env/state-machine/Cargo.toml
@@ -1,12 +1,12 @@
 [package]
 name = "icrc1-test-env-state-machine"
-version = "0.1.2"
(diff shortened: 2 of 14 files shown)


test/env/Changelog.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.12] - 2025-09-08
+### Changed
+- Bump rust to 1.85.0
+
+## [0.1.11] - 2024-01-16
+### Added
+- Duplicate release of 0.1.2
+
```
mbjorkqvist, line 14: Bumping all the crates to `0.1.12`, since this [icrc1-test-env](https://crates.io/crates/icrc1-test-env/versions) crate has a(n accidentally published?) version `0.1.11`. Even though `0.1.2` was published later, the `0.1.11` version is higher, so that's what gets used by most users. To keep all crates at the same level, the next available patch version is `0.1.12`.

test/env/state-machine/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 name = "icrc1-test-env-state-machine"
-version = "0.1.2"
+version = "0.1.12"
 authors = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
 rust-version = { workspace = true }
 repository = { workspace = true }
-description = { workspace = true }
+description = "DEPRECATED: Use icrc1-test-env-pocket-ic"
```
mbjorkqvist, line 9: Marking this crate as deprecated here. The plan is to publish this version, and then in a follow-up PR:

1. Migrate the code to use PocketIC instead of StateMachine
2. Move the code from `test/env/state-machine/` to `test/env/pocket-ic`
3. Rename the crate to `icrc1-test-env-pocket-ic`
4. Bump the crate version in all the crates under `test/` to `0.2.0` and publish them.

PR #203, opened 2025-09-08
feat: FI-1858: Migrate from StateMachine to PocketIC

mbjorkqvist:
Migrate the testing infrastructure to use PocketIC instead of StateMachine.


test/env/pocket-ic/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+fn new_principal(n: u64) -> Principal {
+    let mut bytes = n.to_le_bytes().to_vec();
+    bytes.push(0xfe);
+    bytes.push(0x01);
+    Principal::try_from_slice(&bytes[..]).unwrap()
+}
+
+#[derive(Clone)]
+pub struct SMLedger {
```
maciejdfinity, line 19: I think SM stands for State Machine? In this case we could call it e.g. `PICLedger`? It's not a drop in replacement anyway, since you need to change the parameter of  the `new` call.

